PR #2, opened 2020-06-05
Feature code cov

tialab:
I have added pytest-cov to requirements. This is an extension to pytest (also made by pytest) which generates code coverage reports which will enable us to verify which parts of the code are actually being tested. I have also updated the tox and Travis configuration to output the report to the Travis terminal output and generate an XML which is then uploaded to codecov.

To see an example report you can view the one for this branch here: https://codecov.io/gh/TIA-Lab/tiatoolbox/branch/feature-code-cov.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ script: tox
 
 # Upload test coverage reports (codecov and deepsource)
 after_success:
+# Upload coverage to codecov
 - bash <(curl -s https://codecov.io/bash)
 # Install deepsource CLI
 - curl https://deepsource.io/cli | sh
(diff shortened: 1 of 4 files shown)


tialab:
Closing to reopen under the correct user.

John-P:
Looks like I cannot reopen it as a different user.

.travis.yml
```diff
@@ -13,6 +13,9 @@ install: pip install -U tox-travis
 # Command to run tests, e.g. python setup.py test
 script: tox
 
+after_success:
+- bash <(curl -s https://codecov.io/bash)
```
simongraham, line 17: can we add a comment here to explain what it is doing

tialab:
Thanks for the feedback. I've added some comments to explain what is going on. If everyone is happy then I think this can be merged.

shaneahmed:
> Thanks for the feedback. I've added some comments to explain what is going on. If everyone is happy then I think this can be merged.

Simon had one comment if you have addressed it then this can be merged.

PR #5, opened 2020-06-19
DOC: Create Read The Docs Configuration

John-P:
I have set up a Read the Docs configuration file to see if it can resolve our failing build. Read the Docs was failing because it could not find Numpy after `import tiatoolbox` in the sphinx config file: `docs/conf.py`.

After specifying to use `requiremetns_dev.txt` on the web UI it then failed to load OpenSlide because the shared library / `.so` was missing.

Using a configuration YAML file allows using anaconda (environment.conda.yml) to set up the build environment.


John-P:
Might want to add back pinned / set version of packages now that the docs actually build. I removed them so that conda could be more flexible in solving the environment.

John-P:
What is the status on this? I belive the last I heard this is not working on windows? Has this been resolved and can this be merged or is it an open issue?

shaneahmed:
> What is the status on this @shaneahmed ? I believe the last I heard this is not working on Windows? Has this been resolved and can this be merged or is it an open issue?

This is working on Windows. I have tested this.

PR #10, opened 2020-07-02
Add Feature read a region from whole slide image

shaneahmed:
- Add feature to read a region from whole slide image in WSIReader class.
- Add the option to display or save specific region from whole slide images using command line


tiatoolbox/utils/misc.py
```diff
@@ -75,3 +76,22 @@ def save_yaml(input_dict, output_path="output.yaml"):
     """
     with open(pathlib.Path(output_path), "w") as yaml_file:
         yaml.dump(input_dict, yaml_file)
+
+
+def cv2_imwrite(image_path, cv_im):
```
John-P, line 81: Just wondering if it's better to just call it 'imwrite' rather than 'cv2_imwrite'? So that if later we find a better / faster way to write to file etc we can change the code here and the name would still work e.g. if it wasn't using cv2 anymore. Happy if you want to leave it as is though, just thinking about abstracting away exactly which library is doing the writing.

shaneahmed: I agree I think we should go with `imwrite`. Revised it in this commit https://github.com/TIA-Lab/tiatoolbox/commit/16ad9f93aa21b23344cb6261ff9f037e6d3646ab

requirements.conda.yml
```diff
@@ -4,19 +4,20 @@ channels:
...
-  - pyyaml=5.3.1
-  - requests=2.23.0
+  - python=3.7
+  - setuptools<=45.1.0
+  - Click
+  - cython
+  - h5py
+  - matplotlib
+  - numpy
+  - opencv>=4.0
```
simongraham, line 14: why >= opencv ? What if a later version breaks something?

requirements_dev.txt
```diff
@@ -7,11 +7,15 @@ tox==3.14.0
 coverage==5.1
 Sphinx==1.8.5
 twine==1.14.0
-Click==7.0
-setuptools==45.1.0
+Click>=7.0
```
simongraham, line 10: same as before. Why are we using >=, rather than == ?

tiatoolbox/cli.py
```diff
@@ -2,9 +2,12 @@
 from tiatoolbox import __version__
 from tiatoolbox import dataloader
 from tiatoolbox import utils
+
 import sys
 import click
 import os
+import pathlib
+from PIL import Image
```
simongraham, line 10: Are we using opencv and pillow? Is there a reason for using both libraries?

simongraham: Also, can we fix the spacing in this file.

requirements.conda.yml
shaneahmed: We will specify minimum version number required for the package unless the code breaks, in which case we will create an issue and update the code. Until the code is fixed the version number for that particular package can be pinned.

tiatoolbox/cli.py
shaneahmed: > Are we using opencv and pillow? Is there a reason for using both libraries?

Pillow Image show opens the image in default system browser that's the intention here for command-line.

shaneahmed: > Also, can we fix the spacing in this file.

I don't see any issues with spacing can you point out?

requirements_dev.txt
shaneahmed: We will specify minimum version number required for the package unless the code breaks, in which case we will create an issue and update the code. Until the code is fixed the version number for that particular package can be pinned.

tiatoolbox/cli.py
simongraham: Go ahead if you are fine with it

PR #14, opened 2020-07-10
DOC: Convert Shared RST files to Markdown

tialab:
I have converted the shared files (readme, authors, history) from rst to md and added recommonmark in a new conda dev file so that readthedocs can include them.


tialab:
Close as there was a mistake in direction of PR

PR #21, opened 2020-07-23
API: Restructure WSIReader to allow support for additional formats

shaneahmed:
- Restructure WSIReader as parent class to allow support to read whole slide images in other formats.
- Add OpenSlideWSIReader
- Add tests for OpenSlideWSIReader
- Add docs for OpenSlideWSIReader

### DIFF
--- a/tiatoolbox/cli.py
+++ b/tiatoolbox/cli.py
@@ -53,6 +53,7 @@ def main():
 @click.option(
     "--verbose",
     type=bool,
+    default=True,
     help="Print output, default=True",
 )
 def slide_info(wsi_input, output_dir, file_types, mode, workers=None, verbose=True):
@@ -222,6 +223,12 @@ def slide_thumbnail(wsi_input, output_path, mode):
     help="num of cpu cores to use for multiprocessing, "
     "default=multiprocessing.cpu_count()",
 )
+@click.option(
+    "--verbose",
+    type=bool,
+    default=True,
+    help="Print output, default=True",
+)
 def save_tiles(
     wsi_input,
     output_dir,
@@ -230,6 +237,7 @@ def save_tiles(
     tile_read_size_w,
     tile_read_size_h,
     workers=None,
+    verbose=True,
 ):
     """Displays or saves WSI metadata"""
     file_types = tuple(file_types.split(", "))
@@ -252,6 +260,7 @@ def save_tiles(
         tile_objective_value=tile_objective_value,
         tile_read_size_w=tile_read_size_w,
         tile_read_size_h=tile_read_size_h,
+        verbose=verbose,
         workers=workers,
     )
 
--- a/tiatoolbox/cli.py
+++ b/tiatoolbox/cli.py
@@ -2,6 +2,7 @@
 from tiatoolbox import __version__
 from tiatoolbox import dataloader
 from tiatoolbox import utils
+from tiatoolbox.utils.exceptions import FileNotSupported
 
 import sys
 import click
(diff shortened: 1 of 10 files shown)


tiatoolbox/dataloader/slide_info.py
```diff
@@ -1,20 +1,21 @@
...
-    """Single file run to output or save WSI meta data.
-
-    Multiprocessing uses this function to run slide_info in parallel
+def slide_info(input_path, output_dir=None, verbose=True):
+    """Returns WSI meta data. Multiprocessing decorator runs this function in parallel.
 
     Args:
         input_path (str): Path to whole slide image
         output_dir (str): Path to output directory to save the output
+        verbose(bool): Print output, default=True
```
simongraham, line 16: space after verbose

tiatoolbox/dataloader/slide_info.py
```diff
@@ -34,16 +35,18 @@ def slide_info(input_path, output_dir=None):
...
             input_dir=input_dir, file_name=file_name, output_dir=output_dir
         )
-        info = wsi_reader.slide_info()
+        info = wsi_reader.slide_info
+        if verbose:
+            print(info)
     else:
-        print("File type not supported")
-        info = None
+        raise FileNotSupported
```
simongraham, line 50: Doing it this way will raise a generic error - do we not want to also give the filename for which it failed on?

tiatoolbox/utils/exceptions.py
```diff
@@ -0,0 +1,8 @@
+"""Custom Errors and Exceptions for TIAToolbox"""
+
+
+class FileNotSupported(Exception):
```
simongraham, line 4: instead - word as:  'File format is not supported'

tiatoolbox/dataloader/slide_info.py
shaneahmed: Fixed in this commit https://github.com/TIA-Lab/tiatoolbox/commit/3ac21d1b573dc9f272a8527ef14e866b52c9c631

tiatoolbox/dataloader/slide_info.py
shaneahmed: Updated it in this commit https://github.com/TIA-Lab/tiatoolbox/commit/3ac21d1b573dc9f272a8527ef14e866b52c9c631

tiatoolbox/utils/exceptions.py
shaneahmed: Updated in this commit https://github.com/TIA-Lab/tiatoolbox/commit/dd194c4ee46e2bc4f36c73138f8f03515037b1fd

shaneahmed:
> Overall the PR is good and I like how we subclass the WSIReader. The next step can be to subclass with a JP2Reader. Why do we use double underscore for `__slide_info__`? This is normally reserved for inbuilt python magic methods - we can instead just use double leading underscore to denote a private method. For the exceptions, do you think we should instead simply use a function, with the filename as an argument. This way we can give a more specific error message. Just something to consider.
> 
> Other than that, a couple of very minor things to fix before merging. I will approve, but please look into making those small changes before making the merge. Also, for the module string added to the top of a script, should we not add a space in between the module string and first library import?

Thanks Simon. I have addressed your concerns on ___slide_info__ and error message. Regarding space after docstring at the top of module, let's discuss this on the channel and then make it consistent after adding to OneNote.

tiatoolbox/cli.py
```diff
@@ -119,22 +126,33 @@ def slide_info(wsi_input, output_dir, file_types, mode, workers=None):
 )
 def read_region(wsi_input, region, level, output_path, mode):
     """Reads a region in an whole slide image as specified"""
-    if all(region):
+    if not region:
```
John-P, line 129: Is this checking if that region is not None? I would consider changing it to `if region is not None:` if so as it is more clear.

shaneahmed: Its to check empty tuple "()".

PR #23, opened 2020-08-06
NEW: Feature OmnyxJP2 reader

shaneahmed:
This PR adds functionality to read JP2 images. However, there are some issues which we need to discuss before I can merge it.


vqdang (approved):

tiatoolbox/dataloader/save_tiles.py
```diff
@@ -32,9 +33,10 @@ def save_tiles(
     tile_read_size_h=5000,
     verbose=True,
 ):
-    """Save image tiles for whole slide image. Default file format for tiles is jpg.
-    Multiprocessing decorator runs this function in parallel using the number of
-    specified cpu cores.
+    """Save image tiles for whole slide image.
+    Default file format for tiles is jpg.
+    Multiprocessing decorator runs this function in parallel using
+    the number of specified cpu cores.
```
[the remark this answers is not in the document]
shaneahmed: Well spotted. Fixed it in this commit https://github.com/TIA-Lab/tiatoolbox/commit/372552b0925002b6d4fbf70e391ad7e5487f31ed

tests/test_tiatoolbox.py
```diff
@@ -32,15 +34,15 @@ def _response_ndpi(tmpdir_factory):
     """
     ndpi_file_path = tmpdir_factory.mktemp("data").join("CMU-1.ndpi")
     if not os.path.isfile(ndpi_file_path):
-        print("Downloading NDPI")
+        print("\nDownloading NDPI")
```
John-P, line 37: Curious as to why you are adding an additional line break here (and in other prints), especially as it just used for debugging and won't show up normally.

tests/test_tiatoolbox.py
```diff
@@ -350,7 +439,7 @@ def test_command_line_slide_info(_response_all_wsis):
 
 
 def test_command_line_read_region(_response_ndpi, tmp_path):
-    """Test the Read Region CLI."""
+    """pytest OpenSlide read_region CLI."""
```
John-P, line 442: Not sure about this verbing of pytest

tests/test_tiatoolbox.py
```diff
@@ -451,61 +584,4 @@ def test_command_line_save_tiles(_response_all_wsis, tmp_path):
...
-    wsi_obj = wsireader.OpenSlideWSIReader(input_dir, file_name + ext)
-    meta = wsi_obj.slide_info
-    assert meta.validate()
-
-
-def test_wsimeta_openslidewsireader_svs(_response_svs, tmp_path):
-    input_dir, file_name, ext = utils.misc.split_path_name_ext(str(_response_svs))
-    wsi_obj = wsireader.OpenSlideWSIReader(input_dir, file_name + ext)
-    meta = wsi_obj.slide_info
-    assert meta.validate()
```
John-P, line 511: What happened here?

tiatoolbox/dataloader/wsimeta.py
```diff
@@ -59,7 +59,7 @@ def __init__(
         self.vendor = vendor
         self.mpp = np.array(mpp)
         self.file_path = file_path
-        self.raw = dict(raw) if raw is not None else None
```
John-P, line 62: Just a thought for the future, we should probably provide getter functions (or @properties) for these to stop a user accidentally modifying the original as they are returned by reference.

tiatoolbox/utils/misc.py
```diff
@@ -103,7 +103,8 @@ def imwrite(image_path, img):
     """Write a numpy array to an image
 
     Args:
-        image_path (str, pathlib.Path): file path (including extension) to save image
+        image_path (str, pathlib.Path): file path (including extension)
+        to save image
```
John-P, line 107: Maybe needs indenting

John-P (commented):
Mostly just minor points. Is there a reason that lots of test were removed?

tiatoolbox/dataloader/wsimeta.py
shaneahmed: Can we make it private or read-only?

tests/test_tiatoolbox.py
shaneahmed: I just moved it above in the python API section. This was CLI section

tests/test_tiatoolbox.py
shaneahmed: I will change it to Test and make it consistent

tests/test_tiatoolbox.py
shaneahmed: It was looking weired on my screen so I just added it.

tiatoolbox/utils/misc.py
shaneahmed: Fixed in this commit https://github.com/TIA-Lab/tiatoolbox/commit/25809e3ee8955f97933d290bd7d1b0874f7c6b15

tiatoolbox/dataloader/wsimeta.py
John-P: You cannot have private or read only variables in python. A leading underscore indicate private but does not prevent modification. A double preceding underscore performs 'name mangling'. See https://docs.python.org/3/tutorial/classes.html#private-variables

However, you can make a function with returns a copy and appears as if it is an attribute e.g.:

```python
@property
def raw(self):
    return copy.deepcopy(self.raw)

raw = wsi.raw # Gets a copy
raw["key"] = None # Modifies the copy but original is untouched
```

You can also use this method to memoize costly things e.g. getting slide info here (not particularly costly but an example):

```python
@property
def info(self):
    if not hasattr(self, "_m_info"):
        self._m_info = self._info()
    return self._m_info # or return a copy of to prevent modification
```

PR #24, opened 2020-08-07
EXAMPLE: Add slide info notebook

vqdang:
Just as the title says. It consists of loading WSI and sample code for extracting patches using scipy or the provided functionality from tiatoolbox. And also a more descriptive explanation.


tiatoolbox/dataloader/wsireader.py
```diff
@@ -135,6 +135,12 @@ def save_tiles(self, tile_format=".jpg", verbose=True):
             ...     tile_read_size_w=2000)
             >>> wsi_obj.save_tiles()
 
+        Examples:
+            >>> from tiatoolbox.dataloader import wsireader
+            >>> wsi_obj = wsireader.WSIReader(input_dir="./",
```
shaneahmed, line 140: The input is changed to input_path from JP2 branch. Once that's merged with develop probably you can update it.

shaneahmed (changes requested):
Thanks Dang. This PR looks good for a first attempt. I think it needs a bit of spell checker or editor to fix issues with Grammar and spacing. Please make those changes and I will review the PR again. The notebook needs to be updated.

simongraham (changes requested):
Overall, very good. Please fix the below:

- I would start the notebook with a heading. Of course this uses markdown so use # heading.
- Do not say 'first example'. This may not be the first notebook in the users file directory.
- Instead of hires in comment, use high resolution to be clear.
- Instead of 'load up a small wsi', use 'load a small wsi'. 
- Use WSI. not wsi in the markdown.
- Change to create **an** OpenSlideWSIReader.
- Do not use 'and' before etc. Should read: such as magnification, dimension, etc.
- Use colon after as follows.
- <matplotlib.image.AxesImage at 0x7fb3a2857610> can be supressed by using `plt.show()`
- Hide axes on figures. 
- For the markdown immediately before cell 7, use bullet points for the task.
- Assertion failed at the end - need to fix.

simongraham:
Have pushed to the branch. 

When running locally, I couldn't do the imports as expected. For example, `from tiatoolbox.dataloader.slide_info import slide_info` can't be completed because `tiatoolbox` can't be seen when the root is in the examples folder. 

Also, please update the following markdown text:

Now, we will check the content of the output folder and plot some tiles for visualization. The extracted tiles are be saved under subfolder named by `sample_file_name` under subfolder named `output` under `data_dir`, the subfolder would contain a `.csv` which gives the extracted tiles.

The output location isn't very clear. I would have updated it myself, but wasn't too sure what was meant. Just explicitly give the path, rather then explaining it in with description in between. 

Because of the import issue, the notebook will need to be run again.

shaneahmed (approved):
Please address the comments added to the notebook and merge once second reviewer is happy.

PR #26, opened 2020-08-07
NEW: Metadata Class

John-P:
I have made a few suggested changes to the WSIMeta class, any feedback would be appreciated. Here us a summary fo the changes I suggest.

1. Changed some properties which were a 1-tuple (vendor, mpp_x, mpp_y) in OpenSlideWISReader to not be tuples (they are all returned as strings from OpenSlide I believe). [Commit](https://github.com/TIA-Lab/tiatoolbox/commit/010214bfbe1d46ad48d5517be18bca0764172883#diff-b976b2751d8de6d913055ad9b28c7f2a)
2. Changed the `__slide_info()` method on WSIReader to be public i.e. `slide_info()` and also a property so that it can be accessed via `slide.slide_info`. The WSIMeta object could also be created in the WSIReader `__init__` and asigned which would save re-creating it many times. However, I didn’t want to introduce too many changes at once and making it a property was a simpler fix. [Commit](https://github.com/TIA-Lab/tiatoolbox/commit/8d5082b2653ac6cc198513ea5adff9372746abf0)
3. I removed some things from the metadata which are not really metadata e.g. tile scaling (that is the scaling required to make image tiles at a certain magnification not to be confused with the tiles in the TIFF file) etc. as it is not information about the slide itself.
[Commit](https://github.com/TIA-Lab/tiatoolbox/commit/0d80261704e8cf4342d9c867adb8114819dd2848)
4. I changed some variables from singular to plural to be consistent with OpenSlide etc. e.g. `level_dimension` -> `level_dimensions` as it is a list of values. [Commit 1](https://github.com/TIA-Lab/tiatoolbox/commit/76b1393704612e736a49b64d4736f2ea5ab0e8a3) [Commit 2](https://github.com/TIA-Lab/tiatoolbox/commit/084e0babd3488960b2ad2e884dfae7e18cb2530d) [Commit 3](https://github.com/TIA-Lab/tiatoolbox/commit/1233f951b522948058f0a82e88dc8b9ff7dad5cd)
5. Small spacing and typo fixes.
6. Cast properties to python types. Properties from OpenSlide are returned as string. A user should not have to know to call `float(slide.slide_info.mpp_x)` every time. `slide.slide_info` (also suggest renaming to just info i.e. `slide.info`) should be fine. The raw values can always be accessed via `slide.raw`.
7. Added a copy of the raw metadata from OpenSlide as `slide_info.raw` (cast to a dict to prevent multiprocessing issues with ctypes pointers).
8. Added some data validation e.g. checking that level_count matches up with the length of the level_dimensions and level_downsamples.
9. Added type hints to WSIMeta. This would also make it easier to move to using a [dataclass](https://docs.python.org/3/library/dataclasses.html) for WSIMeta as it is the perfect use case for dataclasses (i.e. classes holding data). It would provide elegant data validation and automatic `to_dict` etc. However, dataclasses were added in python 3.7, and we currently support back to 3.6. An alternative would be to use something like [Pydantic](https://pydantic-docs.helpmanual.io/), however, this would mean adding a dependency.
10. Added some example '@property' calculated values which are convenient e.g. mpp to get a numpy array of mpp in x and y. 

This is a bit of a first pass and could do with some docstring tidying and tests. I wanted to open it as a draft PR but I can't find that option on GitHub anymore.


tiatoolbox/dataloader/wsimeta.py
```diff
@@ -1,36 +1,101 @@
...
+            else None
+        )
+        self.level_downsamples = (
+            [float(x) for x in level_downsamples] if level_downsamples else None
+        )
         self.vendor = vendor
-        self.mpp_x = mpp_x
-        self.mpp_y = mpp_y
+        self.mpp_x = float(mpp_x) if mpp_x else None
+        self.mpp_y = float(mpp_y) if mpp_y else None
```
vqdang, line 43: Should be merged to be a list or numpy array

tiatoolbox/dataloader/wsimeta.py
```diff
@@ -1,36 +1,101 @@
...
+
+        self.validate()
+
+    @property
+    def filepath(self):
+        return self.input_dir / self.file_name
+
+    @property
+    def mpp(self):
+        return np.array([self.mmpx, self.mppy])
```
vqdang, line 54: Refer to previous comment and deal with it there

tiatoolbox/dataloader/wsireader.py
```diff
@@ -320,32 +319,27 @@ def __slide_info(self):
...
         level_count = self.openslide_obj.level_count
         level_dimensions = self.openslide_obj.level_dimensions
         level_downsamples = self.openslide_obj.level_downsamples
         file_name = self.file_name
-        vendor = (self.openslide_obj.properties[openslide.PROPERTY_NAME_VENDOR],)
-        mpp_x = (self.openslide_obj.properties[openslide.PROPERTY_NAME_MPP_X],)
-        mpp_y = (self.openslide_obj.properties[openslide.PROPERTY_NAME_MPP_Y],)
+        vendor = self.openslide_obj.properties[openslide.PROPERTY_NAME_VENDOR]
+        mpp_x = self.openslide_obj.properties[openslide.PROPERTY_NAME_MPP_X]
+        mpp_y = self.openslide_obj.properties[openslide.PROPERTY_NAME_MPP_Y]
```
vqdang, line 329: similar

tiatoolbox/dataloader/wsireader.py
```diff
@@ -320,32 +319,27 @@ def __slide_info(self):
...
-            rescale=rescale,
-            tile_objective_value=tile_objective_value,
-            tile_read_size=tile_read_size.tolist(),
+            slide_dimensions=slide_dimensions,
             level_count=level_count,
             level_dimensions=level_dimensions,
             level_downsamples=level_downsamples,
             vendor=vendor,
             mpp_x=mpp_x,
             mpp_y=mpp_y,
```
vqdang, line 340: Similar

vqdang (approved):

tests/test_tiatoolbox.py
```diff
@@ -454,3 +452,60 @@ def test_command_line_save_tiles(_response_all_wsis, tmp_path):
...
+def test_wsimeta_openslidewsireader_ndpi(_response_ndpi, tmp_path):
+    input_dir, file_name, ext = utils.misc.split_path_name_ext(str(_response_ndpi))
+    wsi_obj = wsireader.OpenSlideWSIReader(input_dir, file_name + ext)
+    meta = wsi_obj.slide_info
+    assert meta.validate()
+
+
+def test_wsimeta_openslidewsireader_svs(_response_svs, tmp_path):
+    input_dir, file_name, ext = utils.misc.split_path_name_ext(str(_response_svs))
+    wsi_obj = wsireader.OpenSlideWSIReader(input_dir, file_name + ext)
```
shaneahmed, line 509: Just to be clear, are we keeping the input as before or changing it to full file path?

shaneahmed (commented):
Overall it looks good. I am glad we invested time on this. 

Just to be clear, are we keeping the input as before or changing it to full file path for OpenSlideWSIReader()?

shaneahmed (approved):
Overall it looks good. I am glad we invested time on this. 

Just to be clear, are we keeping the input as before or changing it to full file path for OpenSlideWSIReader()?

John-P:
@shaneahmed  I have not changed the input to the slide reader classes. This could be added here if you like, or be done in a separate PR. Whatever is more appropriate.

shaneahmed:
> @shaneahmed I have not changed the input to the slide reader classes. This could be added here if you like, or be done in a separate PR. Whatever is more appropriate.

OK, I will change this in JP2 PR as it will be more appropriate to change it their. Thanks for this PR, I will merge it now.

PR #27, opened 2020-08-14
ENH: Tiling Refactor

John-P:
This is a work in progress PR to share what I have been doing

**Here I have tried to split out the save_tiles function in WSI reader into two parts:**
1. A a generator which yields a 3-tuple of (xy index, bounding rect, tile image)
2. A function which calls the tile generator and write to disk (replicating previous functionality).

This preserves the same functionality but make it more modular and the tiling code can be reused e.g. to generate patches. While the tiling was being refactored I also took the opportunity to start adding functions for **calculating an appropriate scaling factor based on objective power or mpp** depending on what the user wants. I think most of the time app will be desired (I always use MPP when generating patches and masks etc) but I am aware that sometimes people want to specify an objective power value.

Additionally **unit tests have been updated for these new functions. In the course of updating the test I have also made a couple of enhancements**:
1. Moved tests to use the  tmp_dir fixture when writing to files so that any created files are automatically removed after the test.
2. Changed the fixtures to downloaded slides with a scope of 'session' so that files are created and torn down once per session.

This PR took quite a while due to many issues which I had running unit tests. Some of which have been addressed by using tmp_dir (files being left behind causing file exists exceptions when a test previously failed etc.) and the session scoping for fixtures. I am, however, still having issues with the unit tests which should be addresses as they are not functioning as they should i.e. a quick sanity check during development.

I am having issues with some tests running for much longer than 10 minutes or hanging indefinitely. Some very simples tests are simply taking far too long to run. I am also finding that some test pass inconsistently. For example this PR passes all tests local if run individually but not if run together which is a red flag. Not sure if they will pass on Travis CI, will have to wait and see. To be clear I am seeing these issues locally on develop and other branches, not just this PR. E.g. https://github.com/TIA-Lab/tiatoolbox/pull/26 has now started timing out on Travis despite no changes to the tests and only a very minor code change (minor change of class attributes) [Travis before change](https://travis-ci.org/github/TIA-Lab/tiatoolbox/builds/715934266), [Travis after](https://travis-ci.org/github/TIA-Lab/tiatoolbox/builds/717847652).

This PR is based on https://github.com/TIA-Lab/tiatoolbox/pull/26 due to some changes there which this depends on.


John-P:
Closing due to recent PRs superseding some part of making this incompatible. Recommend basing a future tiling refactor on changes being introduced in #40.

PR #32, opened 2020-09-09
DEP: Depreciate support for multiprocessing

shaneahmed:
- Depreciate support for multiprocessing 
- Make multiprocessing support a separate repository


John-P:
Works reliably on both of my machines which previously had stuck tests 👍

shaneahmed:
> Works on both of my machines which previously had stuck tests 👍 I sometimes get a file exists assertion failure. I think the save_tiles tests is leaving some files behind. This should be fixed when tmp_dir is added back.

This is fixed in https://github.com/TIA-Lab/tiatoolbox/pull/29

PR #49, opened 2020-11-04
NEW: Feature pixel patch generation

shaneahmed:
- Add `get_patch_extractor()` which returns `PointsPatchExtractor`, `FixedWindowPatchExtractor` and `VariableWindowPatchExtractor`
- Add `PointsPatchExtractor` returns patches which can be used by classification models. Take `csv`, `json` or `pd.DataFrame` and returns patches corresponding to each pixel location.
- Add support for ensemble classification.


tiatoolbox/tools/patchextraction.py
```diff
@@ -0,0 +1,361 @@
...
+        num_examples_per_patch(int): Number of examples per patch for ensemble
+         classification.
+
+    """
+
+    def __init__(
+        self,
+        input_image,
+        labels,
+        img_patch_h=224,
```
shaneahmed, line 213: Convert to tuple

tiatoolbox/tools/patchextraction.py
```diff
@@ -0,0 +1,361 @@
...
+    if not isinstance(input_image, np.ndarray):
+
+        if isinstance(input_image, pathlib.Path):
+            input_image = str(input_image)
+
+        if isinstance(input_image, str):
+            _, _, suffix = split_path_name_ext(input_image)
+
+            if suffix in (".jpg", ".png"):
+                input_image = imread(input_image)
```
shaneahmed, line 317: Wrapper class

shaneahmed: Added in this commit https://github.com/TIA-Lab/tiatoolbox/commit/ae7bb3079412829b3b8a79c8126a202c4170c446

tiatoolbox/tools/patchextraction.py
shaneahmed: Fixed in this commit https://github.com/TIA-Lab/tiatoolbox/commit/9f1865f8bf2eb6b49eb5b19b485e3f522a524b31

tiatoolbox/tools/patchextraction.py
```diff
@@ -0,0 +1,382 @@
...
+        Calling this function for :obj:`PointsPatchExtractor` will raise an error. This
+        overrides the merge_patches function in the base class :obj:`PatchExtractor`
+
+        """
+        raise MethodNotSupported(
+            message="Merge patches not supported for PointsPatchExtractor"
+        )
+
+
+def read_point_annotations(input_table):
```
shaneahmed, line 306: Move to misc.py

shaneahmed: Added in this commit https://github.com/TIA-Lab/tiatoolbox/commit/b72516bc3f1436a317256989fd5f4f93b410ed91

tiatoolbox/tools/patchextraction.py
```diff
@@ -0,0 +1,330 @@
...
+    # def extract_patches(self):
+    #     """Extract patches from an image using locations provided by labels data.
+    #
+    #     Args:
+    #         labels (str, ndarray):
+    #
+    #     Returns:
+    #         img_patches (ndarray): extracted image patches of size NxHxWxD.
+    #     """
+    #     raise NotImplementedError
```
John-P, line 92: Should probably remove commented code if it's not being used

John-P:
Here is a link to the documentation for this PR: https://tia-toolbox.readthedocs.io/en/feature-pixel-patch-gen/

tiatoolbox/tools/patchextraction.py
mostafajahanifar: Agreed. I will check them, but maybe Shan need them.

tiatoolbox/tools/patchextraction.py
```diff
@@ -0,0 +1,330 @@
...
+
+        Args:
+            patches: patch-level predictions
+
+        Returns:
+            image: merged prediction
+        """
+        raise NotImplementedError
+
+
```
mostafajahanifar, line 105: If merge_patches is a way to accumulate classification results for example, I am not sure if it is the right place for it. I think it is better to keep this class to patch extraction only. Also, based on input arguments, we do not accept any input to merge here.

tiatoolbox/tools/patchextraction.py
```diff
@@ -0,0 +1,330 @@
...
+    Attributes:
+        stride(Tuple of int): stride in (x, y) direction for patch extraction.
+
+    """
+
+    def __init__(
+        self,
+        input_img,
+        patch_size,
+        pad_size=0,
```
mostafajahanifar, line 121: Where this `pad_size` is being used? if the goal is to pad image to extract patches beyond image size, I think the new PR #78 would resolve that and we do not need to have this argument here. Unless, we want user to decide have padded patches or valid sized patches in which case this should be binary flag.

tiatoolbox/tools/patchextraction.py
```diff
@@ -0,0 +1,330 @@
...
+        return self[n]
+
+    def __getitem__(self, item):
+        if type(item) is not int:
+            raise TypeError("Index should be an integer.")
+
+        if item >= self.labels.shape[0]:
+            raise IndexError
+
+        x, y, _ = self.labels.values[item, :]
```
mostafajahanifar, line 229: I guess it is important to return the classification label alongside the patch as it is necessary to have that data for classification, suggest using something like this:
`x, y, cls = self.labels.values[item, :]`
and in the end return like this:
`return data, cls`

Maybe even (x, y) information would be useful. Should we consider them as optional output?

mostafajahanifar: Also, if we want to support point lists with two columns, we should adjust this part accordingly.

tiatoolbox/utils/misc.py
```diff
@@ -330,3 +332,52 @@ def contrast_enhancer(img, low_p=2, high_p=98):
...
+        >>> from tiatoolbox.utils.misc import read_point_annotations
+        >>> labels = read_point_annotations('./annotations.csv')
+
+    """
+    if isinstance(input_table, (str, pathlib.Path)):
+        _, _, suffix = split_path_name_ext(input_table)
+
+        if suffix == ".npy":
+            out_table = np.load(input_table)
+            out_table = pd.DataFrame(out_table, columns=["x", "y", "class"])
```
mostafajahanifar, line 359: what if point list is not stored in (x, y, class) format?
I guess it is ok to limit the user to follow (x, y) order, but I propose we should support 2-columned point lists as well (when there is no classification information).

tiatoolbox/tools/patchextraction.py
```diff
@@ -0,0 +1,330 @@
...
+        labels(pd.DataFrame): A table containing location and/or type of patch.
+        num_examples_per_patch(int): Number of examples per patch for ensemble
+         classification.
+
+    """
+
+    def __init__(
+        self,
+        input_img,
+        labels,
```
mostafajahanifar, line 200: I suggest using name `locations_list`, `annotations_list`, or 'points_list` instead of label. 

Also, it is good to accept the desired points list as a python `class 'list'`.

tiatoolbox/tools/patchextraction.py
shaneahmed: Removed in this commit https://github.com/TIA-Lab/tiatoolbox/commit/a53769828a7db4d9d696a030f1854382df86756c

tiatoolbox/tools/patchextraction.py
shaneahmed: We may not need it as we can now read beyond boundaries but this PR is focused on points patch extraction and setting up the code for other patch extraction methods.

tiatoolbox/tools/patchextraction.py
shaneahmed: you are right but currently our focus is on inference so we don't need this yet. But this will be added when we add code for training models.

tiatoolbox/tools/patchextraction.py
shaneahmed: Fixed in this commit https://github.com/TIA-Lab/tiatoolbox/commit/0ad8ea23bbb8348cc1627fdd5aed5d30d838216c

tiatoolbox/utils/misc.py
shaneahmed: Fixed in this commit https://github.com/TIA-Lab/tiatoolbox/commit/7b071c51df267efb36e1c204060d017442902d2b

tiatoolbox/tools/patchextraction.py
shaneahmed: `merge_patches` will be required in window patch extraction. As the method needs to be implemented in the base class thats why it's here.

tiatoolbox/utils/misc.py
```diff
@@ -335,6 +337,59 @@ def contrast_enhancer(img, low_p=2, high_p=98):
...
+            out_table = np.load(input_table)
+            if out_table.shape[1] == 2:
+                out_table = pd.DataFrame(out_table, columns=["x", "y"])
+                out_table["class"] = None
+            else:
+                out_table = pd.DataFrame(out_table, columns=["x", "y", "class"])
+
+        elif suffix == ".csv":
+            out_table = pd.read_csv(input_table)
+            if "x" not in out_table.columns:
```
mostafajahanifar, line 370: what if csv file has header? 
I guess we can force the user to use a csv file without header, but it is nicer to consider cases that have a first-line header.

tiatoolbox/utils/misc.py
```diff
@@ -335,6 +337,59 @@ def contrast_enhancer(img, low_p=2, high_p=98):
...
+                out_table = pd.read_csv(
+                    input_table, header=None, names=["x", "y", "class"]
+                )
+
+        elif suffix == ".json":
+            out_table = pd.read_json(input_table)
```
mostafajahanifar, line 376: what if the JSON file had only x, y information?
Should add the class column as well.

tiatoolbox/utils/misc.py
```diff
@@ -335,6 +337,59 @@ def contrast_enhancer(img, low_p=2, high_p=98):
...
+
+        else:
+            raise FileNotSupported("Filetype not supported.")
+
+    elif isinstance(input_table, np.ndarray):
+        out_table = pd.DataFrame(input_table, columns=["x", "y", "class"])
```
mostafajahanifar, line 382: what about numpy arrays with two columns? something like this is preferred:

```
out_table = pd.DataFrame(input_table, columns=["x", "y", "class"])
if input_table.shape[1]==2:
    out_table["class"] = None
```

tiatoolbox/utils/misc.py
```diff
@@ -335,6 +337,59 @@ def contrast_enhancer(img, low_p=2, high_p=98):
     return np.uint8(img_out)
 
 
+def read_point_annotations(input_table):
+    """Read annotations as pandas DataFrame.
```
mostafajahanifar, line 341: I suggest the `out_table` should have three column in all cases, no matter even if the input has only 2 columns. This way, the `PointsPatchExtractor` can be written more consistently regarding the point class output.

tiatoolbox/utils/misc.py
```diff
@@ -335,6 +337,59 @@ def contrast_enhancer(img, low_p=2, high_p=98):
...
+            out_table = pd.read_json(input_table)
+
+        else:
+            raise FileNotSupported("Filetype not supported.")
+
+    elif isinstance(input_table, np.ndarray):
+        out_table = pd.DataFrame(input_table, columns=["x", "y", "class"])
+
+    elif isinstance(input_table, pd.DataFrame):
+        out_table = input_table
```
mostafajahanifar, line 385: Here as well, `out_table` should be checked if it has three columns.

tiatoolbox/tools/patchextraction.py
```diff
@@ -0,0 +1,268 @@
...
+        y = y - int((self.patch_size[0] - 1) / 2)
+
+        data = self.wsi.read_rect(
+            location=(int(x), int(y)),
+            size=self.patch_size,
+            resolution=self.resolution,
+            units=self.units,
+        )
+
+        return data
```
mostafajahanifar, line 228: here check if `c is not None`, then `return data, c`, otherwise `return data`.

tiatoolbox/tools/patchextraction.py
```diff
@@ -0,0 +1,268 @@
...
+    """
+
+    def __init__(
+        self,
+        input_img,
+        locations_list,
+        patch_size=224,
+        resolution=0,
+        units="level",
+        num_examples_per_patch=9,
```
mostafajahanifar, line 189: I suggest the default value for `num_examples_per_patch` be 1, as model ensembling should be optional.

tiatoolbox/tools/patchextraction.py
```diff
@@ -0,0 +1,268 @@
...
+        return self[n]
+
+    def __getitem__(self, item):
+        if type(item) is not int:
+            raise TypeError("Index should be an integer.")
+
+        if item >= self.locations_list.shape[0]:
+            raise IndexError
+
+        x, y, _ = self.locations_list.values[item, :]
```
mostafajahanifar, line 216: when adjusted `read_point_annotations` to always return class as well (None if class is absent), this line can be written like this, to support point class in the output as well (if applicable):
`x, y, c = self.locations_list.values[item, :]`

tiatoolbox/utils/misc.py
shaneahmed: the header should be `x`, `y` and `class` as in the documentation.

mostafajahanifar: Sorry, I get it wrong. this part of code supports both header and headerless csv files. But, the two column situation is not covered here as well.  I can work on it if you want.

tiatoolbox/utils/misc.py
shaneahmed: Added in this commit https://github.com/TIA-Lab/tiatoolbox/commit/478ec912426334b82d74e5a8e4c01494f62779ac

tiatoolbox/utils/misc.py
shaneahmed: Added in this commit https://github.com/TIA-Lab/tiatoolbox/commit/478ec912426334b82d74e5a8e4c01494f62779ac

tiatoolbox/utils/misc.py
shaneahmed: Added in this commit https://github.com/TIA-Lab/tiatoolbox/commit/478ec912426334b82d74e5a8e4c01494f62779ac

tiatoolbox/utils/misc.py
shaneahmed: Added in this commit https://github.com/TIA-Lab/tiatoolbox/commit/478ec912426334b82d74e5a8e4c01494f62779ac

tiatoolbox/tools/patchextraction.py
shaneahmed: Agreed, this was to allow for augmentation as in SCCNN paper. But current implementation allows only 1 example.

tiatoolbox/tools/patchextraction.py
shaneahmed: This may generate an error due to unused variable. When we use `c` in future then we can add it. We are not using it currently therefore there is no need to assign a value to it.

tiatoolbox/tools/patchextraction.py
shaneahmed: `c` will be added in future. Let's keep it to what we need now, we may need to change the output depending on usage.

tiatoolbox/tools/patchextraction.py
mostafajahanifar: Yes you're right.
Actually, I wrote a code that extracts c for all scenarios, but I didn't push it because at my first push I realized that there are many changes required to be done in the test section to cover that and I was not sure if even you want that to be added or not!

tiatoolbox/tools/patchextraction.py
shaneahmed: Fixed in this commit https://github.com/TIA-Lab/tiatoolbox/commit/8e3bbe7dc1e1807dd0b7fc4bd558ddd1bd56c585

tiatoolbox/utils/misc.py
```diff
@@ -338,6 +337,77 @@ def contrast_enhancer(img, low_p=2, high_p=98):
...
+    if isinstance(input_table, (str, pathlib.Path)):
+        _, _, suffix = split_path_name_ext(input_table)
+
+        if suffix == ".npy":
+            out_table = np.load(input_table)
+            if out_table.shape[1] == 2:
+                out_table = pd.DataFrame(out_table, columns=["x", "y"])
+                out_table["class"] = None
+            else:
+                out_table = pd.DataFrame(out_table, columns=["x", "y", "class"])
```
vqdang, line 366: Need strict enforcing  shape[1] == 3 here else pd will throw out error, which may be uninformative for end user, also may want to raise error or warning shape[1] > 3

tiatoolbox/utils/misc.py
vqdang: Re the csv, same as above, error will likely occur here due to wrong csv format and we would not be able to deal with it. I think we should wrap it in a `try` block and only accept strict `csv` format and mention it so within the doc.

tiatoolbox/utils/misc.py
```diff
@@ -338,6 +337,77 @@ def contrast_enhancer(img, low_p=2, high_p=98):
...
+                    input_table,
+                    header=None,
+                    names=["x", "y", "class"],
+                    sep=None,
+                    engine="python",
+                )
+            if out_table.shape[1] == 2:
+                out_table["class"] = None
+
+        elif suffix == ".json":
```
vqdang, line 381: Same as above, need to wrap within `try` block. One more thing, I'm not sure what should be the format within `json` here, do you want it as follow or sthg else?
```
{
'x' : [0,1,2,3],
'y' : [1,2,3,4],
'class' : [0, 0, 1, 1]
}
```

vqdang (changes requested):

vqdang:
@shaneahmed Minor question aside the actual PR code. Is it normal to include test data within the commit ? It bloats and clusters up the #files abit.

tiatoolbox/utils/misc.py
shaneahmed: It will be as you mentioned above. I generated json using pd.to_json so I am assuming this to be default format.

tiatoolbox/utils/misc.py
shaneahmed: Resolved in this PR https://github.com/TIA-Lab/tiatoolbox/commit/854f4549144f618309dd275f1026120f41f41888

tiatoolbox/utils/misc.py
shaneahmed: Resolved in this commit as discussed https://github.com/TIA-Lab/tiatoolbox/commit/854f4549144f618309dd275f1026120f41f41888

vqdang (approved):

mostafajahanifar (approved):

PR #50, opened 2020-11-05
BUG: Fix Reading Thumbnail In WSIReader

John-P:
Fixes the `get_thumbnail` function in `WSIReader`. It now uses `read_bounds` with the origin and the slide dimensions as the bounds.


shaneahmed (approved):
Please merge it with develop if the tests pass.

PR #52, opened 2020-11-06
TST: Split tests into multiple files

shaneahmed:
- Split tests into multiple files to reduce clutter and avoid accidental edits.


John-P (approved):
All looks good to me 👍

PR #53, opened 2020-11-10
NEW: Fallback To Calculate MPP From Standard TIFF Tags

John-P:
This is a draft PR proposing adding some code to calculate the microns-per-picel (MPP) based off of standard TIFF tags.

This is useful in the case where slides and been modified and converted to standard tiles tiffs and use regular TIFF tags instead of putting metadata in the description field (or when created directly as regular tiled TIFFs). In this case openslide does not find the MPP. Examples of such images can be seen in the [PANDA Kaggle challenge dataset](https://www.kaggle.com/c/prostate-cancer-grade-assessment/data).

As always the MPP is not completely reliable. In the future, it would be good to have some method to sanity check the values e.g. to make sure a sensible range or train a model to predict from the image if missing etc.

Also fixes a couple of bugs where the reader would raise an exception and fail if metadata keys could not be found. Now they are simply set to None if not found.

### DIFF
--- a/tiatoolbox/dataloader/wsireader.py
+++ b/tiatoolbox/dataloader/wsireader.py
@@ -837,13 +837,16 @@ class OpenSlideWSIReader(WSIReader):
 
         """
         props = self.openslide_wsi.properties
-        objective_power = int(props[openslide.PROPERTY_NAME_OBJECTIVE_POWER])
+        if openslide.PROPERTY_NAME_OBJECTIVE_POWER in props:
+            objective_power = int(props[openslide.PROPERTY_NAME_OBJECTIVE_POWER])
+        else:
+            objective_power = None
 
         slide_dimensions = self.openslide_wsi.level_dimensions[0]
         level_count = self.openslide_wsi.level_count
         level_dimensions = self.openslide_wsi.level_dimensions
         level_downsamples = self.openslide_wsi.level_downsamples
-        vendor = props[openslide.PROPERTY_NAME_VENDOR]
+        vendor = props.get(openslide.PROPERTY_NAME_VENDOR)
 
         # Find microns per pixel (mpp)
         # Initialise to None (value if cannot be determined)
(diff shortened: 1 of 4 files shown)


tiatoolbox/dataloader/wsireader.py
```diff
@@ -836,18 +836,42 @@ def info(self):
...
 
         """
-        objective_power = np.int(
-            self.openslide_wsi.properties[openslide.PROPERTY_NAME_OBJECTIVE_POWER]
-        )
+        props = self.openslide_wsi.properties
+        if openslide.PROPERTY_NAME_OBJECTIVE_POWER in props:
+            objective_power = int(props[openslide.PROPERTY_NAME_OBJECTIVE_POWER])
+        else:
+            objective_power = None
```
shaneahmed, line 843: Please guess objective from mpp if objective value not available but with a warning to the user.

shaneahmed (approved):
Great! we can now move towards supporting tiff. 
Please address the comment and merge when you are ready.

vqdang (approved):

tiatoolbox/dataloader/wsireader.py
```diff
@@ -836,18 +836,42 @@ def info(self):
...
+
+        # Find microns per pixel (mpp)
+        # Initialise to None (value if cannot be determined)
+        mpp = None
+        # Check OpenSlide for mpp metadata first
+        try:
+            mpp_x = props[openslide.PROPERTY_NAME_MPP_X]
+            mpp_y = props[openslide.PROPERTY_NAME_MPP_Y]
+            mpp = [mpp_x, mpp_y]
+        # Fallback to TIFF resolution units and convert to mpp
```
vqdang, line 859: Add a warning message saying the mpp retrieved from tiff metadat.

vqdang (approved):
Need to add another warning message.

tiatoolbox/dataloader/wsireader.py
John-P: Just added in the latest commit

tiatoolbox/dataloader/wsireader.py
John-P: I've added a function to approximate and some tests.

PR #55, opened 2020-11-15
BUG: Fix 2 Bugs In read_region

John-P:
The read bounds function for openslide was using the incorrect location for levels other than level 0. This has been fixed in the included commit. I am currently writing a test to detect and avoid this error in future.

### DIFF
--- a/tests/test_wsireader.py
+++ b/tests/test_wsireader.py
@@ -1,3 +1,4 @@
+from tiatoolbox.utils.transforms import imresize
 from tiatoolbox.dataloader import wsireader
 from tiatoolbox import utils
 from tiatoolbox import cli
@@ -6,6 +7,7 @@ import pytest
 from pytest import approx
 import pathlib
 import numpy as np
+import cv2
 from click.testing import CliRunner
 
 # -------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


vqdang (approved):

John-P:
I found and fixed a second bug while writing a test which could detect the first bug. This was in finding the best level to use when reading. If the read operation requested a resolution greater than baseline (max resolution level), then it would try to read at the highest level (lowest resolution) instead of the lowest level (the highest resolution) and upscaling by interpolation.

This has now been fixed also.

John-P:
Summary of PR (draft for merge commit message):

Fix Two Bugs In WSI Read Methods

Fix two bugs in read methods, one in `read_region` and the other in `_find_optimal_level_and_downsample`.

1. Change OpenSlideWSIReader `read_bounds` to use coordinates in baseline coordinates instead of in level coordinates when calling OpenSlide `read_region method.
2. Add check for requested resolution > baseline in _find_optimal_level_and_downsample. Return level 0 if requested resolution is higher than baseline (level 0). Currently, `np.argmax` returns 0 in this case (all relative level scales > 1). However, it is performed on a reversed list and therefore the read level becomes level_count - 1 which is incorrect.

shaneahmed (approved):

PR #58, opened 2020-12-03
BUG : Directory input for stainnorm operation

Srijay-lab:
Fix a bug in taking directory input for stainnorm operation. All files should be read from the input text if it is a directory instead of returning an empty list.

Note: It is a critical bug fix as it changed the syntax of default file types, which is a common input for all commands and that's why affects all other operations.


shaneahmed (approved):
Thanks Srijay. Looks fine to me.

PR #60, opened 2020-12-10
TST: Add Constants For Test Regions

John-P:
I propose defining some constants for WSI regions which contain some tissue for use in the tests rather than putting numbers in each individual test.

To keep the PR simple for initial review and discussion I have just committed the change adding the constants. Of course the existing tests would also have to be updated to use them.


tests/test_wsireader.py
```diff
@@ -9,6 +9,22 @@
 import cv2
 from click.testing import CliRunner
 
+# -------------------------------------------------------------------------------------
+# Constants
+# -------------------------------------------------------------------------------------
+
+NDPI_TEST_TISSUE_BOUNDS = (30400, 11810, 30912, 12322)
```
shaneahmed, line 16: Looks fine to me, but I cannot see if you are using these in the file somewhere.

shaneahmed (commented):
Happy with the proposed changes but I can't see where you are using these.

John-P:
I have adjusted tests to use these region constants.

John-P:
# PR Summary

- Add region constants for WSIReader tests.
- Swap literals in tests for defined constants.

PR #62, opened 2020-12-10
NEW: Add Safe & Sub-Pixel Image Reads

John-P:
Add ability to read images with sub-pixel coordinates (if upsampling while reading) and also to safely read areas outside the image (similar to OpenSlide read_region) without an index error. By default, it is zero padded i.e. black but can be changed to reflection, symmetric or other padding easily.

This PR additionally adds a small utility for finding the size of a bounds tuple: `bounds2size`.

The added functionality is split into two parts:

`safe_padded_read`
--------------------------

This allows reads outside the image region (0 - width and 0-height). It takes integer bounds but they may be any integers (negative or outside the image bounds).

out_of_bounds_demo

`sub_pixel_read`
---------------------------------

This allows for sub-pixel reads. You can supply a set of bounds as floats and an output size. The region will be read using the fractional bounds and upscaled using a given interpolation method (defaults to cubic).

### Modularity & Potential Use Beyond Numpy Images
Internally this method calls safe_padded_read. However, this can be swapped out for any method via the read_func kwarg. read_func is given takes integer bounds, although they may be outside the image bounds. This can easily be replaced e.g. with some small wrapper around OpenSlide read_region or glymur read. This makes it more flexible and able to be used in other contexts.

### Interpolation Padding
This also accounts for boundary effects resulting from interpolation. By default, padding is added to the image before interpolating and then removed after. This is because some interpolation methods such as cubic and lanczos have a region of support over a few pixels which could extend outside the final visible region. This can be disabled with a key word argument.

### Fancy Rounding
There is also some fancy rounding in there to ensure the output image size is consistent with the requested output size. This may also be of use or inspiration for other read functions to help ensure this consistency. E.g. there is often an 1-off error with glymur reads.

### Illustration of various options

sub_pixel_demo

The difference that extra padding for interpolation makes is very subtle. Here is the comparison zoomed in:

sub_pixel_demo_interp_padding

## Tests
Tests have added a couple of tests: one for known troublesome parameters and another fuzz test. The fuzz test uses 10,000 sets of random parameters to check that these functions do not throw errors and produce consistent results i.e. output image size matches requested size.


John-P:
Docs building at https://tia-toolbox.readthedocs.io/en/feature-subpixel-numpy-reads/usage.html#module-tiatoolbox.utils

tests/test_utils.py
```diff
@@ -58,6 +59,167 @@ def test_mpp2common_objective_power(_sample_svs):
...
+    bounds = (1.5, 1, 5, 5)
+    with pytest.raises(ValueError):
+        utils.image.safe_padded_read(data, bounds)
+
+
+def test_safe_padded_read_negative_padding():
+    """Test safe_padded_read with non-integer bounds."""
+    data = np.zeros((16, 16))
+
+    bounds = (1, 1, 5, 5)
```
shaneahmed, line 75: Aren't these integer bounds?

tiatoolbox/utils/image.py
```diff
@@ -0,0 +1,370 @@
...
+
+    Args:
+        image (:class:`numpy.ndarray`):
+            Image to read from.
+        bounds (tuple(float)):
+            Bounds of the image to read in
+            (left, top, right, bottom) format.
+        output_size (tuple(int)):
+            The desired output size.
+        padding (int, tuple(int)):
```
shaneahmed, line 160: Please add default value.

shaneahmed (commented):
Looks good to me. There are a couple of comments added. 

- I was wondering if we should move imread and imwrite from misc to image.read and image.write?
- Also as image is usually used as a variable should we use it as a module name?

tests/test_utils.py
John-P: I'm not sure what you are asking here. This is a test for a function which only accepts integer bounds. Are you thinking of the sub-pixel read function? This is the safe padded read function which the sub-pixel read function calls internally with the bounding integer coordinates.

John-P:
> Looks good to me. There are a couple of comments added.
> 
>     * I was wondering if we should move imread and imwrite from misc to image.read and image.write?
> 
>     * Also as image is usually used as a variable should we use it as a module name?

- RE moving read and write to 'image'. I am OK with this. Could also go in and 'io' module like in scikit-image.
- RE using 'image' as a module name. I think this is ok e.g. scipy uses ndimage for n-dimensional images. However, open to suggestions of alternative names / locations for this.

tests/test_utils.py
shaneahmed: It looks fine to me. If I remember correctly, these values were not integers that's why I wrote the comment. It looks fine to me now.

shaneahmed:
> > Looks good to me. There are a couple of comments added.
> > ```
> > * I was wondering if we should move imread and imwrite from misc to image.read and image.write?
> > 
> > * Also as image is usually used as a variable should we use it as a module name?
> > ```
> 
> * RE moving read and write to 'image'. I am OK with this. Could also go in and 'io' module like in scikit-image.
> * RE using 'image' as a module name. I think this is OK e.g. scipy uses 'ndimage' as the name for an n-dimensional image ~module~ package. However, open to suggestions of alternative names / locations for this. I just can't think of a better alternative myself at the moment.

ok let's discuss these in the meeting on Friday.

shaneahmed (approved):

PR #64, opened 2020-12-13
DEV: Update HTML Coverage Recipe

John-P:
Change recipe in Makefile for coverage to use pytest-cov instead of coverage. Pytest-cov supports multiprocessing and a few extra bits. Also added pytest-cov to dev requirements so that it will not throw and error when using the recipe. Also for consistency across testing (travis / tox is using pytest-cov and the env is slightly different when using coverage).

To generate test coverage locally run the make recipe:

```bash
$ make coverage
```

This will run the tests and collect coverage before opening the index in a web browser:

image

Clicking through to a file will show line-by-line coverage:

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,9 +57,7 @@ test-all: ## run tests on every Python version with tox
 	tox
 
 coverage: ## check code coverage quickly with the default Python
-	coverage run --source tiatoolbox -m pytest
-	coverage report -m
-	coverage html
+	pytest  --cov=tiatoolbox --cov-report=term --cov-report=html
 	$(BROWSER) htmlcov/index.html
 
 docs: ## generate Sphinx HTML documentation, including API docs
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -57,9 +57,7 @@ test-all: ## run tests on every Python version with tox
 	tox
 
 coverage: ## check code coverage quickly with the default Python
-	coverage run --source tiatoolbox -m pytest
-	coverage report -m
-	coverage html
+	pytest  --cov=tiatoolbox --cov-report=term --cov-report=html
```
shaneahmed, line 60: Do we need to exclude any files from coverage?

shaneahmed (approved):
Looks good. Just one comment if we need to ignore any files for coverage? Otherwise you can merge it with develop.

John-P:
> 
> 
> Looks good. Just one comment if we need to ignore any files for coverage? Otherwise you can merge it with develop.

I do not think so. That would probably be done via the setup.cnf file and I would probably make that a separate PR as the makefile does not affect this.

PR #67, opened 2020-12-18
ENH: Cache Slide Info Property

John-P:
- Add caching to slide info property. This is done by checking if a private `self._m_info` exists and returning it if so, otherwise `self._info` is called to create the info for the first time (or to force regenerating) and the result is assigned to `self._m_info`. This could in future be made much simpler with the `functools.cached_property` decorator in Python 3.8+.
- Add `copy.deepcopy` when retuning info, i.e. return a copy. This prevents accidentally mutating the slide info object.
- Add cast around openslide property (ctype property map) to make it a python dictionary instead. This means that the resulting `WSIMeta` object does not contain any ctypes pointers and is therefore picklable.
- Add a setter for slide info for explicitly updating the info `WSIMeta` object.
- Update tests.

### DIFF
--- a/tiatoolbox/dataloader/wsireader.py
+++ b/tiatoolbox/dataloader/wsireader.py
@@ -24,6 +24,7 @@ from tiatoolbox.dataloader.wsimeta import WSIMeta
 
 import pathlib
 import warnings
+import copy
 import numpy as np
 import openslide
 import glymur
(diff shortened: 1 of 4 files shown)


tiatoolbox/dataloader/wsireader.py
```diff
@@ -1077,14 +1096,14 @@ def read_rect(self, location, size, resolution=1.0, units="baseline"):
         )
 
         im_region = self.img[
-            level_location[1]:level_location[1] + baseline_read_size[1],
-            level_location[0]:level_location[0] + baseline_read_size[0],
+            level_location[1] : level_location[1] + baseline_read_size[1],
```
shaneahmed, line 1099: DeepSource doesn't like it.

John-P: Interesting because these lines are not modified by this PR

John-P: Ah yes, I see what it is. This is black and deepsource style guides conflicting. Style issues are non-blocking on deepsource though.

John-P: I just merged changed from develop and there were some more non black compliant tests but that is now fixed.

tiatoolbox/dataloader/wsireader.py
```diff
@@ -59,15 +60,33 @@ def __init__(
...
+
+    @info.setter
+    def info(self, meta):
+        """WSI metadata setter.
 
         Args:
-            self (WSIReader):
+            meta (WSIMeta): Metadata object.
+        """
+        self._m_info = meta
```
shaneahmed, line 83: Shouldn't this be tested?

John-P: This is now covered is a test

shaneahmed (approved):

PR #78, opened 2021-01-08
ENH: Allow VirtualWSIReader To Read Past Boundaries

John-P:
- Make `VirtualWSIReader` use `utils.image.safe_padded_read` to read past image boundaries.
- Add stride arguments to `utils.image.safe_padded_read` and `utils.image.sub_pixel_read` to support reading multi-resolution arrays (such as glymur images) more efficiently with padding beyond their bounds and sub-pixel coordinates respectively.
- Make `OmnyxJP2Reader` use `utils.image.sub_pixel_read` to allow sub-pixel reads.
- Add `info` kwarg to `VirtualWSIReader` to pass in metadata (currently only being set to attribute but in preparation of implementing virtual pyramids).
- Add new transform to convert a location and size to bounds `utils.transforms.locsize2bounds`.
- Add round-trip fuzz test for `bounds2locsize` and `locsize2bounds`.
- Replace depreciated `glymur.read` with slicing reads.
- Add fuzz (randomised) test for `utils.image.safe_padded_read`.
- General tidying.


John-P:
Doc building here(linking to added util function): https://tia-toolbox.readthedocs.io/en/enhance-virtualwsireader-padded-read/usage.html#tiatoolbox.utils.misc.conv_out_size

tiatoolbox/utils/transforms.py
```diff
@@ -164,6 +164,25 @@ def bounds2size(bounds, origin="upper"):
             Defaults to upper.
     """
     left, top, right, bottom = bounds
+    origin = origin.lower()
     if origin == "upper":
         return np.array([right - left, bottom - top])
-    return np.array([right - left, top - bottom])
+    if origin == "lower":
```
shaneahmed, line 170: I couldn't get this one. What do you mean by origin="lower"? and why would we need top-bottom?

John-P:
Deepsource raising a style error here but looks ok to me and might be false positive. Perhaps the raw string is confusing deepsource https://deepsource.io/gh/TIA-Lab/tiatoolbox/run/183519cb-9b8c-4118-b7af-0b1271cd988c/python/FLK-D207

John-P:
Removed some unused variable to appease deepsource and also tidied the path logic in cli methods. Might want to add a test to check for slide_info cli file output in correct location. I am unsure where it is supposed to be output here so input would be helpful.

tiatoolbox/utils/transforms.py
John-P: This is if the original point is inverted e.g. matplotlib uses lower left by default. This defaults to upper as it is the convention for graphics contexts and images.

tests/test_utils.py
```diff
@@ -90,6 +90,24 @@ def test_safe_padded_read_padding_formats():
...
+        utils.image.safe_padded_read(data, bounds, padding=(1, 1, 1))
+
+
+def test_safe_padded_read_stride_shape():
+    """Test safe_padded_read for padding size."""
+    data = np.zeros((16, 16))
+
+    bounds = (1.5, 1, 5, 5)
+    with pytest.raises(ValueError):
+        utils.image.safe_padded_read(data, bounds, stride=(1, 1, 1))
```
John-P, line 108: Just spotted this, the stride argument is incorrect. It is a 3-tuple but it is unpacked to x and y stride value internally. This should probably error. Could also allow 3-tuple for striding channel although I cannot see that being used enough to worry about it for now.

shaneahmed:
> Removed some unused variables to appease deepsource and also tidied the path logic in CLI methods. Might want to add a test to check for slide_info CLI file output in correct location. I am unsure where it is supposed to be output here so your input @shaneahmed on this would be helpful.

I think all the default output paths in cli should have consistent policy. I can see you have kept it to input.parent/{meta, im_region, ...} which is fine with me. As long as we are consistent it should be ok.

tests/test_utils.py
shaneahmed: I am not sure if I got your question right. This was supposed to be incorrect to generate an error, right?

John-P:
I have started getting a segmentation fault in `test_stainnorm` from line 73 in `get_concentrations` (OLS when calling`np.linalg`). Can anyone else replicate this issue? Currently waiting to see if it is replicated on Travis.

tiatoolbox/dataloader/wsireader.py
```diff
@@ -891,7 +882,9 @@ def _info(self):
         # Fallback to calculating objective power from mpp
         if objective_power is None:
             if mpp is not None:
-                objective_power = misc.mpp2common_objective_power(np.mean(mpp))
+                objective_power = utils.misc.mpp2common_objective_power(
+                    float(np.mean(mpp))
+                )
```
vqdang, line 887: Mention this in the doc.

tiatoolbox/dataloader/wsireader.py
```diff
@@ -1059,14 +1068,28 @@ class VirtualWSIReader(WSIReader):
...
-    def __init__(self, input_img):
+    def __init__(
+        self,
+        input_img,
+        baseline_size=None,
+        mpp=None,
+        power=None,
+        level_downsamples=None,
+        level_dimensions=None,
+    ):
```
vqdang, line 1079: why provide these manually ? we have the common meta class, should receive that one as input. The virtual will likely need to match the source WSI anyways, and the source has its meta readied.

tiatoolbox/dataloader/wsireader.py
```diff
@@ -1059,14 +1068,28 @@ class VirtualWSIReader(WSIReader):
...
             self.img = input_img
         else:
-            self.img = misc.imread(self.input_path)
+            self.img = utils.misc.imread(self.input_path)
+
+        self._slide_dimensions = baseline_size
+        self._objective_power = power
+        self._mpp = mpp
+        self._level_downsamples = level_downsamples
+        self._level_dimensions = level_dimensions
```
vqdang, line 1092: Same as above, best to route it to same meta class.

tiatoolbox/dataloader/wsireader.py
```diff
@@ -1111,17 +1134,18 @@ def read_rect(self, location, size, resolution=1.0, units="baseline"):
...
-        im_region = transforms.imresize(
+        im_region = utils.transforms.imresize(
             img=im_region, scale_factor=post_read_scale, output_size=size
         )
 
-        im_region = transforms.background_composite(image=im_region)
+        im_region = utils.transforms.background_composite(image=im_region)
         return im_region
 
     def read_bounds(self, bounds, resolution=1.0, units="baseline"):
```
vqdang, line 1151: Add kwargs to the input so that padding kwargs can be pass in.

tiatoolbox/dataloader/wsireader.py
```diff
@@ -1131,16 +1155,18 @@ def read_bounds(self, bounds, resolution=1.0, units="baseline"):
...
             units=units,
         )
-        start_x, start_y, end_x, end_y = bounds
         stride = 2 ** read_level
 
-        im_region = self.img[start_y:end_y:stride, start_x:end_x:stride]
+        im_region = utils.image.safe_padded_read(
+            self.img, bounds, pad_mode="constant", constant_values=255
+        )
+        im_region = utils.transforms.imresize(img=im_region, scale_factor=stride)
```
vqdang, line 1163: Related to above comment about kwargs

tiatoolbox/utils/misc.py
```diff
@@ -313,14 +313,17 @@ def contrast_enhancer(img, low_p=2, high_p=98):
...
     Examples:
         >>> from tiatoolbox import utils
         >>> img = utils.misc.contrast_enhancer(img, low_p=2, high_p=98)
 
     """
     # check if image is not uint8
     if not img.dtype == np.uint8:
-        raise Exception("Image should be RGB uint8.")
+        raise AssertionError("Image should be RGB uint8.")
     img_out = img.copy()
```
vqdang, line 327: By the way of accessing, not RGB but any np.uint8. Change the message.

tiatoolbox/utils/transforms.py
vqdang, line 170: Change the return behavior to match with locsize2bounds. Ideally locsize2bound <=> bound2locsize

tests/test_utils.py
```diff
@@ -90,6 +90,24 @@ def test_safe_padded_read_padding_formats():
...
+        utils.image.safe_padded_read(data, bounds, padding=(1, 1, 1))
+
+
+def test_safe_padded_read_stride_shape():
+    """Test safe_padded_read for padding size."""
+    data = np.zeros((16, 16))
+
+    bounds = (1, 1, 5, 5)
+    with pytest.raises(ValueError):
+        utils.image.safe_padded_read(data, bounds, stride=(1, 1, 1))
```
vqdang, line 108: Can we make this into randomsize read test along boundary of 4 edges ? Also, I don't see you put size check on the output. If this funtion fails, error wont be detected in wsi read region because we have auto resize following it.

vqdang (changes requested):
Need additional test /verification for sanity.

tiatoolbox/dataloader/wsireader.py
vqdang: We may also need a func to allow fast swapping metadata.

tiatoolbox/dataloader/wsireader.py
```diff
@@ -53,14 +53,12 @@ class WSIReader:
...
-    def __init__(
-        self,
-        input_img,
-    ):
+    def __init__(self, input_img):
         if isinstance(input_img, np.ndarray):
             self.input_path = None
         else:
             self.input_path = pathlib.Path(input_img)
+        self._m_info = None
```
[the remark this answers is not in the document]
John-P: Short for memoized info.

tiatoolbox/dataloader/wsireader.py
John-P: These were defined as arguments because it seemed a bit clunky to have to define a metadata object before calling the init but I can certainly see the sense.

RE the fast swapping. You can reassign the info with another, the `@info.setter` allows swapping out via simple assignment e.g. `wsi.info = WSIMeta(...)`.

tests/test_utils.py
John-P: Oh I didn't write these tests but sure can do that.

tiatoolbox/dataloader/wsireader.py
John-P: I may do this in a separate PR to reduce the complexity of this one as it is already getting quite complex.

tiatoolbox/dataloader/wsireader.py
John-P: Same as above, I will make a new PR for this.

John-P:
@vqdang I have tried to address all of your comments. I have suggested opening a second PR for passing padding arguments through on read_region and read_bounds as this PR is getting a bit long and complex.

vqdang (approved):

shaneahmed (approved):
Thanks John. I think it is ready to merge.

PR #84, opened 2021-02-05
DOC: Fix Output Path file_path in Example of slide_info

Srijay-lab:
Change the example given for slide_info() so as to make it work on windows.


John-P (approved):
Looks good. To me. Just need to update the PR title to the correct format and write the merge commit message Edit: I would also remove reference to Windows as this would also error on Ubuntu etc.

PR #112, opened 2021-05-19
EG: Tune all examples to fit into master branch

mostafajahanifar:
The title says the content. Reordering the example notebooks, their link, and content as for the sake of master classes.


shaneahmed (commented):
@mostafajahanifar please can you update the intro in 01-example?

shaneahmed (approved):
Thanks @mostafajahanifar

PR #117, opened 2021-05-24
HOTFIX: Update model URLs to fix tests 

simongraham:
- [x] Updated URLs to models in line with update in the backend
- [x] Add `ConnectionError` if URL not found to give more informative error


shaneahmed (approved):
Thanks @simongraham 
Looks fine to me.

PR #120, opened 2021-05-28
ENH: Add functionality to read through mask in `patchextraction`

mostafajahanifar:
This PR is focused on `patchextraction` functionality to clean up some part, change the API, and adding new features to it.

The `SlidingWindowPatchExtraction` class will accept the `input_mask` argument to extract patches based on the regions that are highlighted in the `input_mask`. The `input_mask` can be a path to the desired mask mask, a numpy array, `VirtualWSIReader`, or `'auto'`. In case of `'auto'`, a tissue mask is automatically generated for the `input_image` using tiatoolbox `TissueMasker` functionality if the `input_image` is a real WSI, otherwise, no mask will be considered for patch extraction.

- [x] Rewrite the `slidingwindow` patch extraction class to use `get_coordinates` static method for patch bounding box generation.
- [x] Add mask-based patch extraction functionality based on the `filter_coordinates` static method.
- [x] Add tests for coverage.

### DIFF
--- a/tiatoolbox/tools/patchextraction.py
+++ b/tiatoolbox/tools/patchextraction.py
@@ -36,6 +36,13 @@ class PatchExtractor(ABC):
         input_img(str, pathlib.Path, :class:`numpy.ndarray`): input image for
           patch extraction.
         patch_size(int or tuple(int)): patch size tuple (width, height).
+        input_mask(str, pathlib.Path, :class:`numpy.ndarray`, or :obj:WSIReader):
+          input mask that is used for position filtering when extracting patches
+          i.e., patches will only be extracted based on the highlighted regions in
+          the input_mask. input_mask can be either path to the mask, a numpy
+          array, :class:`VirtualWSIReader`, or 'auto`. In case of 'auto', a
+          tissue mask is generated for the input_image using tiatoolbox
+          :class:`TissueMasker` functioncality.
         resolution (int or float or tuple of float): resolution at
           which to read the image, default = 0. Either a single
           number or a sequence of two numbers for x and y are
@@ -76,6 +83,7 @@ class PatchExtractor(ABC):
         self,
         input_img,
         patch_size,
+        input_mask=None,
         resolution=0,
         units="level",
         pad_mode="constant",
@@ -93,6 +101,14 @@ class PatchExtractor(ABC):
         self.wsi = wsireader.get_wsireader(input_img=input_img)
         self.locations_df = None
         self.stride = None
+        if input_mask is None:
+            self.mask = None
+        elif input_mask == "auto":
+            self.mask = self.wsi.tissue_mask(resolution=1.25, units="power")
+        else:
+            self.mask = wsireader.VirtualWSIReader(
+                self.input_mask, info=self.wsi.info, mode="bool"
+            )
 
     def __iter__(self):
         self.n = 0
@@ -304,6 +320,7 @@ class SlidingWindowPatchExtractor(PatchExtractor):
         self,
         input_img,
         patch_size,
+        input_mask=None,
         resolution=0,
         units="level",
         stride=None,
@@ -312,6 +329,7 @@ class SlidingWindowPatchExtractor(PatchExtractor):
     ):
         super().__init__(
             input_img=input_img,
+            input_mask=input_mask,
             patch_size=patch_size,
             resolution=resolution,
             units=units,
--- a/tiatoolbox/tools/patchextraction.py
+++ b/tiatoolbox/tools/patchextraction.py
@@ -218,48 +218,6 @@ class SlidingWindowPatchExtractor(PatchExtractor):
         raise NotImplementedError
 
 
-class VariableWindowPatchExtractor(PatchExtractor):
-    """Extract and merge patches using variable sized windows for images and labels.
-
-    Args:
-        stride(tuple(int)): stride in (x, y) direction for patch extraction.
-        label_patch_size(tuple(int)): network output label (width, height).
-
-    Attributes:
-        stride(tuple(int)): stride in (x, y) direction for patch extraction.
-        label_patch_size(tuple(int)): network output label (width, height).
-
-    """
-
-    def __init__(
-        self,
-        input_img,
-        patch_size,
-        resolution=0,
-        units="level",
-        stride=(1, 1),
-        pad_mode="constant",
-        pad_constant_values=0,
-        label_patch_size=None,
-    ):
-        super().__init__(
-            input_img=input_img,
-            patch_size=patch_size,
-            resolution=resolution,
-            units=units,
-            pad_mode=pad_mode,
-            pad_constant_values=pad_constant_values,
-        )
-        self.stride = stride
-        self.label_patch_size = label_patch_size
-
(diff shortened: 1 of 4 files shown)


tiatoolbox/tools/patchextraction.py
```diff
@@ -329,10 +262,8 @@ def get_patch_extractor(method_name, **kwargs):
     """
     if method_name.lower() == "point":
         patch_extractor = PointsPatchExtractor(**kwargs)
-    elif method_name.lower() == "fixedwindow":
-        patch_extractor = FixedWindowPatchExtractor(**kwargs)
-    elif method_name.lower() == "variablewindow":
-        patch_extractor = VariableWindowPatchExtractor(**kwargs)
+    elif method_name.lower() == "sliding":
```
shaneahmed, line 265: Can we call it `slidingwindow` instead of `sliding` ?

tiatoolbox/tools/patchextraction.py
```diff
@@ -92,7 +110,20 @@ def __init__(
         self.n = 0
         self.wsi = wsireader.get_wsireader(input_img=input_img)
         self.locations_df = None
+        self.coord_list = None
         self.stride = None
+        if input_mask is None:
+            self.mask = None
+        elif input_mask == "auto":
```
shaneahmed, line 117: can we specify `otsu` and `morphological` as are the options in tissue mask?
change `auto`  to `default` where tissue mask can use default mode.

vqdang (approved):

tiatoolbox/tools/patchextraction.py
```diff
@@ -21,7 +21,8 @@
 """This file defines patch extraction methods for deep learning models."""
 from abc import ABC
 import numpy as np
-import math
+
+# import math
```
shaneahmed, line 25: you can remove it if not needed, no need to comment.

tiatoolbox/tools/patchextraction.py
```diff
@@ -36,6 +37,13 @@ class PatchExtractor(ABC):
         input_img(str, pathlib.Path, :class:`numpy.ndarray`): input image for
           patch extraction.
         patch_size(int or tuple(int)): patch size tuple (width, height).
+        input_mask(str, pathlib.Path, :class:`numpy.ndarray`, or :obj:WSIReader):
+          input mask that is used for position filtering when extracting patches
+          i.e., patches will only be extracted based on the highlighted regions in
+          the input_mask. input_mask can be either path to the mask, a numpy
+          array, :class:`VirtualWSIReader`, or 'auto`. In case of 'auto', a
```
shaneahmed, line 44: adjust according to change in option to add `otsu` and `morphological`

tiatoolbox/tools/patchextraction.py
```diff
@@ -54,14 +62,22 @@ class PatchExtractor(ABC):
         pad_constant_values (int or tuple(int)): Values to use with
           constant padding. Defaults to 0. See :func:`numpy.pad` for
           more.
+        within_bound (bool): whether to extract patches beyond the
```
shaneahmed, line 65: what if it is False.
change if yes to if True

tiatoolbox/tools/patchextraction.py
```diff
@@ -54,14 +62,22 @@ class PatchExtractor(ABC):
...
     Attributes:
         wsi(WSIReader): input image for patch extraction of type :obj:`WSIReader`.
         patch_size(tuple(int)): patch size tuple (width, height).
         resolution(tuple(int)): resolution at which to read the image.
         units (str): the units of resolution.
         n(int): current state of the iterator.
-        locations_df(pd.DataFrame): A table containing location and/or type of patch.
+        locations_df(pd.DataFrame): A table containing location and/or type of patces
+          in `(x_start, y_start, class)` format.
+        coord_list(:class:`numpy.ndarray`): An array containing coordinates of patches
```
shaneahmed, line 78: why we need this separately?

tiatoolbox/tools/patchextraction.py
```diff
@@ -146,21 +177,163 @@ def _generate_location_df(self):
...
+        return np.array(flag_list)
+
+    @staticmethod
+    def get_coordinates(
+        image_shape=None,
+        patch_shape=None,
+        stride_shape=None,
+        within_bound=True,
+    ):
+        """Calculate patch tiling coordinates.
```
shaneahmed, line 269: how is it different from _locations_df?

shaneahmed (commented):
Thanks @mostafajahanifar 
Please can you address the comments before we can merge?

tiatoolbox/tools/patchextraction.py
mostafajahanifar: resolved in d1bfe24cdc0100a5b1a855be87ab8adb100a7926

tiatoolbox/tools/patchextraction.py
mostafajahanifar: `locations_df` contains only starting `x,y` positions while `coord_list` bounding box information to be used for read bound. I kept both at the moment as `locations_df` might have been used somewhere else.

tiatoolbox/tools/patchextraction.py
mostafajahanifar: `get_coordinates` is static function that is scale/resolution agnostic and can be used anywhere on demand. While `_generate_location_df` is class specific function that uses the former and other function and tricks to calculate resolution and region specific locations for patch extraction. Having these two functions separated allow us to expand this part more easily in the future.

tiatoolbox/tools/patchextraction.py
mostafajahanifar: Yeah it's good idea to include `otsu` and `motphological` method as input. It's better than `auto` and if we are specifying the methods, the `auto` option is no longer valid. However, the default value for this argument should remain `None` as it's needed to distinguish between mask-based or non-mask-based patch extraction.

tiatoolbox/tools/patchextraction.py
mostafajahanifar:
```
within_bound (bool): Whether to include the patches on the right and bottom
                margins of mother image. If `True`, the patches that their location
                exceeds the `image_shape` would be neglected. Otherwise, those patches
                would be extracted with `Reader` function and appropriate padding.
```

I have changed the class docstrings accordingly.

tiatoolbox/tools/patchextraction.py
mostafajahanifar: resolved in 137e25c758490f665c967a81f0263c34f5d17677

tiatoolbox/tools/patchextraction.py
mostafajahanifar: resolved in 137e25c758490f665c967a81f0263c34f5d17677

tiatoolbox/tools/patchextraction.py
mostafajahanifar: resolved in 137e25c758490f665c967a81f0263c34f5d17677

shaneahmed (approved):
Thanks @mostafajahanifar  and @vqdang

PR #122, opened 2021-06-04
DOC: Fix examples in module docstring

Srijay-lab:
Modify the documentation for following:

- Use get_wsireader() instead of wsireader() in the [documentation](https://tia-toolbox.readthedocs.io/en/latest/usage.html)

- Replace file_name with file_path in the documentation of example of slide_info() function

- Include numpy import statement in the example of tiatoolbox.utils.transforms.background_composite


shaneahmed (approved):
Thanks @Srijay-lab 
It looks good.

rajasaad (approved):

PR #141, opened 2021-08-31
DEV: Extra Pre-commit Checks

John-P:
I have compiled many extra checks for pre-commit, some of which are the same as those used by deepsource. Proposed new additions to pre-commit:

- isort: Sort and group imports (stdlibrary, 3rd party, current package)
- flake8-isort: Check import ordering
- flake8-comprehensions: Suggestions for better comprehensions
- flake8-mutable: Check for mutable default arguments
- flake8-use-fstring: Use of f-strings vs old style format strings
- flake8-simplify: Suggestions to simplify code
- pep8-naming: Check PEP8 class naming
- flake8-eradicate: Find dead/commented out code
- flake8-builtins: Check for built-ins being used as variables
- flake8-cognitive-complexity: Check max function complexity (set in setup.cfg)
- flake8-expression-complexity: Check max expression complexity (set in setup.cfg)
- flake8-return: Check return statements (no redundant assignments)
- flake8-spellcheck: Spelling checker (comments only, new words in whitelist.txt)
- flake8-pytest-style: Check against pytest style guide
- flake8-fixme: Check for FIXME, TODO, and XXX left in comments
- flake8-sql: Check SQL statement style
- nitpick: Check that black, flake8, isort have compaible options
- flake8-logging-format: Validate (lack of) logging format strings
- flake8-pie: Misc. linting rules

Of course we don't have to use all of these but I went through a bunch of them and tried them out. These were the ones that looked helpful and did not have any obvious annoying bugs or conflicts.

Some of these could be added to travis also if that would be useful.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,6 +18,10 @@ universal = 1
 exclude = docs, *__init__*
 max-line-length = 88
 extend-ignore = E203 # For black compatibility
+spellcheck-targets = comments
+dictionaries = en_US,python,technical
+max-cognitive-complexity = 7
+max-expression-complexity = 7
 
 [aliases]
 test = pytest
@@ -25,3 +29,9 @@ test = pytest
 [tool:pytest]
 collect_ignore = ['setup.py']
 
+[isort]
+profile = black
+multi_line_output = 3
+skip_gitignore = True
+line_length = 88
+filter_files = True
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,10 @@
 repos:
+  - repo: https://github.com/pycqa/isort
+    rev: 5.8.0
+    hooks:
+      - id: isort
   - repo: https://github.com/psf/black
-    rev: 20.8b1 # Replace by any tag/version: https://github.com/psf/black/tags
+    rev: 20.8b1 # Replace with any tag/version: https://github.com/psf/black/tags
     hooks:
       - id: black
         language_version: python3 # Should be a command that runs python3.6+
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -18,10 +18,20 @@ universal = 1
...
 
 [tool:pytest]
 collect_ignore = ['setup.py']
 
+[isort]
+profile = black
+multi_line_output = 3
+skip_gitignore = True
+line_length = 88
+filter_files = True
```
shaneahmed, line 37: empty line?

shaneahmed (commented):
Thanks @John-P I am in favour of these pre-commits. My only question, can we skip some errors from these packages to allow commit in case the programmer would like to force skip?

setup.cfg
John-P: empty line??

John-P:
> Thanks @John-P I am in favour of these pre-commits. My only question, can we skip some errors from these packages to allow commit in case the programmer would like to force skip?

So, Looking into the flake8 and pre-commit configs/options. There are a few things we could do:
1. Run a subset of checks on commit and more checks on push. You can customise which commands run on any of: commit, push, writing commit message, checkout, merge, etc.
2. You can use `$ commit --no-verify` to by pass the hooks for a single commit.
3. We could also split the checks into two groups, one of which will always return true (and set verbose true to still print, as passing tests don't print by default) and therefore not ever cause the action to block but will still print out a message.

Example Pre-Commit Hook That Will Always Print No Mater The Exit Code:

```yaml
  - repo: local
    hooks:
      - id: say-hello
        name: Say Hello
        description: Just prints a message.
        entry: python -c "print('Hello there!')"
        language: python
        verbose: true
```

While looking at the documentation, I also noticed that pre-commit can now auto set up when the repo is cloned. This is probably suited for a different PR though.

shaneahmed (approved):

PR #144, opened 2021-09-10
DOC: Improve docstrings for readthedocs

simongraham:
Cleaned up docstrings so that the output of readthedocs was correct.

For this, it was important to use 2 spaces, rather than an indent with 4 spaces when going to a continuing text onto a new line in the docstring.

The `rst` file for semantic segmentation is not yet done - this needs to be completed.

- Some docstrings have a colon. For example, https://github.com/TIA-Lab/tiatoolbox/blob/develop/tiatoolbox/tools/patchextraction.py#L42 - this needs to be addressed otherwise readthedocs will have an undesirable output.


docs/usage.rst
```diff
@@ -128,6 +128,25 @@ Classification
 .. automodule:: tiatoolbox.models.classification.pretrained_info
     :members:
 
+------------
```
shaneahmed, line 131: Is it possible to decouple this PR from segmentation models? To help with the release with patch predictor?

You can add these changes to semantic segmentation PR.

tests/test_models_semantic_segmentation.py
```diff
@@ -0,0 +1,451 @@
+import copy
```
shaneahmed, line 1: Same as above, if we can add these changes to the semantic segmentation that would be great. Just to help with the release on patch predictor.

tiatoolbox/models/segmentation/__init__.py
```diff
@@ -0,0 +1,28 @@
+# ***** BEGIN GPL LICENSE BLOCK *****
```
shaneahmed, line 1: same as above

tiatoolbox/models/segmentation/semantic_segmentor.py
```diff
@@ -0,0 +1,984 @@
+# ***** BEGIN GPL LICENSE BLOCK *****
```
shaneahmed, line 1: same as above

shaneahmed (commented):
Thanks @simongraham 
Overall it looks fine to me. Would it be possible to decouple this PR from the segmentation PR? Just to help with the release? You can merge the changes related to segmentation branch on its PR.

tiatoolbox/models/classification/patch_predictor.py
```diff
@@ -152,11 +152,9 @@ class CNNPatchPredictor:
         >>> data = [img1, img2]
```
shaneahmed, line 152: The Attributes list is not complete. I think you are missing `pretrained_models` and `pretrained_weights`.

tiatoolbox/models/classification/patch_predictor.py
```diff
@@ -152,11 +152,9 @@ class CNNPatchPredictor:
...
         >>> predictor = CNNPatchPredictor(pretrained_model="resnet18-kather100k")
         >>> output = predictor.predict(data, mode='patch')
-
         >>> data = np.array([img1, img2])
         >>> predictor = CNNPatchPredictor(pretrained_model="resnet18-kather100k")
         >>> output = predictor.predict(data, mode='patch')
-
         >>> wsi_file = 'path/wsi.svs'
```
shaneahmed, line 158: Add an example with all the available options for `CNNPatchPredictor`.

simongraham: Addressed

shaneahmed (approved):
Thanks @simongraham Readthedocs look fine to me now.

PR #155, opened 2021-10-13
NEW: Add SlideGraph Construction Method In Tools

John-P:
Add classes and functions under `tiatoolbox.tools.graph` to enable construction of graphs in a format which can be used with PyG (PyTorch Geometric).


tests/test_graph.py
```diff
@@ -0,0 +1,145 @@
...
+    # Expect 1 connection, symmetrical so dividing by 2
+    assert np.sum(adjacency_matrix) / 2 == 1
+
+
+def test_affinity_to_edge_index_fuzz_output_shape():
+    """Fuzz test that output shape is 2xM for affinity_to_edge.
+
+    Output is 2xM, where M is the number of edges in the graph, i.e.
+    the number of connections between nodes with a value > threshold.
+    """
```
shaneahmed, line 74: new line.

tiatoolbox/tools/graph.py
```diff
@@ -0,0 +1,275 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# The Original Code is Copyright (C) 2021, TIALab, University of Warwick
+# All rights reserved.
+# ***** END GPL LICENSE BLOCK *****
+
+"""Functions to help with constructing graphs."""
```
shaneahmed, line 21: if it is just torch based graphs then please mention it.

tiatoolbox/tools/graph.py
```diff
@@ -0,0 +1,275 @@
...
+        # Initialise similarity of coordinate i vs all coordinates to 1
+        # (most un-similar).
+        i_vs_all_similarities = np.ones(len(points))
+        # Set the neighbours similarity to calculated values (similarity/fd)
+        i_vs_all_similarities[neighbour_indexes[i]] = neighbour_similarities
+        i_vs_all_similarities = i_vs_all_similarities[i + 1 :]
+        condensed_distance_matrix[
+            index : index + len(i_vs_all_similarities)
+        ] = i_vs_all_similarities
+        index = index + len(i_vs_all_similarities)
```
vqdang, line 151: How long would this take? It's better to support
- pre-generated distance matrix NxM, this is what sklearn does for Agglomerative Clustering
- directly using pdist when NxM is small
- and this last resort above

The main concern will usually the size of the matrix, but looping like this wont utilize numpy parallel very well. Additionally, this seems to be 2 ways linkage between points, so I think there are some wasted processing here.

At the moment this is ok as this is more like a prototype rather than an optimal implementation, you may want to revisit this later though.

tiatoolbox/tools/graph.py
```diff
@@ -0,0 +1,275 @@
...
+    label: Optional[int] = None,
+    lambda_d: Number = 3.0e-3,
+    lambda_f: Number = 1.0e-3,
+    lambda_h: Number = 0.8,
+    connectivity_distance: Number = 4000,
+    neighbour_search_radius: Number = 2000,
+    feature_range_thresh: Number = 1e-4,
+) -> Dict[str, ArrayLike]:
+    """Build a graph via hybrid clustering in spatial and feature space.
+
```
vqdang, line 47: Could you clarify this for me? It seems the clustering is for 2D density spatial location using the node features. Thus,
it is different from DBSCAN. You may have a better word for [spatial] here vs DBSCAN. On another note, feature space is still 'spatial' somewhat...

tiatoolbox/tools/graph.py
```diff
@@ -0,0 +1,275 @@
...
+    Returns:
+        dict: A dictionary defining a graph for serialisation (e.g.
+        JSON) or converting into a torch-geometric Data object.
+            - :class:`numpy.ndarray` - x:
+                Features of each node (mean of features in a cluster).
+            - :class:`numpy.ndarray` - edge_index:
+                Edge index matrix defining connectivity.
+            - :py:obj:`Number` - y:
+                The label of the graph.
```
vqdang, line 96: Return a set of nodes where each is a 2D spatial cluster from the input points?

The return/example description doesn't describe well what each value means, and the initial intro also does not.

tiatoolbox/tools/graph.py
```diff
@@ -0,0 +1,275 @@
...
+        neighbours = traingle_neighbours[index]
+        neighbours = np.array(list(neighbours), dtype=int)
+        kdtree = cKDTree(points[neighbours, :])
+        nearby_neighbours = kdtree.query_ball_point(
+            x=points[index],
+            r=dthresh,
+        )
+        neighbours = neighbours[nearby_neighbours]
+        adjacency[index, neighbours] = 1.0
+        adjacency[neighbours, index] = 1.0
```
vqdang, line 236: This part doesn't seem to be Delaunay to me. What is kdtree doing here? @wenqi006

vqdang (changes requested):
Need clarification for functionalities. IMO there are some questionable points about the implementation but that can be dealt with later.

tiatoolbox/tools/graph.py
John-P: The KDTree is used to only connect neighbours within a certain distance. This can be done without a KDTree but for many points it if faster to use a KDTree. You make a good point, though. This step could possibly be skipped if the distance threshold is set to 0 or a negative number to mean no threshold on maximum distance to connect.

tiatoolbox/tools/graph.py
wenqi006: The clustering here is based on the similarity achieved from the spatial distance (euclidean distance) and feature difference between node pairs. Therefore we name it as hybrid clustering in spatial and feature space. But as you said that feature space is still 'spatial', what about change it as 'hybrid clustering based on euclidean distance and feature similarity'?

John-P: It is essentially just a weighted combination of distances in different domains or  reference frames. Here, we define it for spatial (i.e. x, y locations within a slide) and features (e.g. resnet features) as this is what SlideGraph uses. However, this function could be made more generic by accepting a list of sets of coordinates and a list of corresponding weightings. E.g. if you wanted to weight with three domains/reference frames such as: location in WSI, H&E features, IHC features etc.

```python
def hybird_clustered_graph(
  point_sets: ArrayLike,
  weights: List[Number],
  connectivity_distance: Number ... etc
)
```

tiatoolbox/tools/graph.py
wenqi006: Also, this is not a 2-way linkage between points. No direction between two nodes. For example, we have a list of nodes [1,2,3,4,5]. Then the length of this matrix 'condensed_distance_matrix' is 10 (5x(5-1)/2).

vqdang: @wenqi006 Not the best way to express non-duplicated pairing but ok for now. I didnt realized that.
@John-P It would be best if this is under a class, as it seems to be dedicated to 1 form of spatial clustering. tbhn, likely wont be in time for the draft so just go with this for now

wenqi006:
Have changed sklearn.neighbors.KDTree into scipy.spatial.cKDTree and made the KDTree package consistent.

Have tested that the current results (neighbour_distances_singlepoint, neighbour_indexes_singlepoint, condensed_distance_matrix) using cKDTree are same to the ones from the previous code which was using sklearn.neighbors.KDTree

tiatoolbox/tools/graph.py
John-P: @vqdang @wenqi It just occurred to me that the number of neighbours to check here is quite small. We could replace the kdtree with a simple list enumeration e.g.:

```python
nearby_neighbours = [n for n in neighbours if np.linalg.norm(n, points[index]) < dthresh]
```

vqdang:
@John-P  May be of interest as sklearn seems to beat scipy (ckdtree version) on kdtree query, and internally both of them use C.
https://stackoverflow.com/questions/30447355/speed-of-k-nearest-neighbour-build-search-with-scikit-learn-and-scipy

shaneahmed (commented):
Thanks @John-P

shaneahmed (approved):
Thanks @John-P

PR #157, opened 2021-10-14
DOC: New Format For Generating Doc Files

vqdang:
Mainly concern with making the doc looks better.
- [x] Automate API generation instead of manual linkage.
- [x] Add back intro page or swap header names to make it easier for new python user navigation.


shaneahmed (commented):
readthedocs fails as below:

The conflict is caused by:
    The user requested torch==1.9.1
    torchvision 0.9.1 depends on torch==1.8.1

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

requirements.txt
```diff
@@ -5,6 +5,7 @@ glymur
 imagecodecs
 jupyterlab
 matplotlib
+nbsphinx
```
shaneahmed, line 8: do we need this outside dev environment?

requirements.txt
```diff
@@ -16,7 +17,8 @@ scikit-image==0.18.3
 scikit-learn>=0.23.2
 shapely
 sphinx==4.1.2
-shapely
+sphinx_gallery
```
shaneahmed, line 20: do we need this outside dev environment?

requirements.txt
```diff
@@ -16,7 +17,8 @@ scikit-image==0.18.3
...
+sphinx_toolbox
```
shaneahmed, line 21: do we need this outside dev environment?

requirements.txt
shaneahmed: Resolved in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/157/commits/06b2d319969a284b19c4c631c72a11fd5c2dc632

requirements.txt
shaneahmed: Resolved in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/157/commits/06b2d319969a284b19c4c631c72a11fd5c2dc632

requirements.txt
shaneahmed: Resolved in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/157/commits/06b2d319969a284b19c4c631c72a11fd5c2dc632

shaneahmed (approved):
Thanks @vqdang

PR #160, opened 2021-10-15
BUG: Fix ioconfig bug for semantic segmentation `tile` mode

vqdang:
Bugs when using `tile` mode, consider

```python
semantic_segmentor = SemanticSegmentor(
    batch_size=4, pretrained_model="fcn-tissue_mask"
)
```

This will work
``` python
config = dict(
    input_resolutions=[{'units': 'baseline', 'resolution': 1.0}],
    output_resolutions=[{'units': 'baseline', 'resolution': 1.0}],
    patch_input_shape=[1024, 1024],
    patch_output_shape=[512, 512],
    stride_shape=[256, 256],
    save_resolution={'units': 'baseline', 'resolution': 0.5},
)
semantic_segmentor.predict(
    [mini_wsi_jpg],
    mode="tile",
    ioconfig=config,
    on_gpu=ON_GPU,
    crash_on_exception=True,
    save_dir=f"{save_dir}/raw/",
)
```
But this will crash even though `ioconfig` is supposed to have been generated with `pretrained_model` . 

**Reason**: Internally `tile` config has been designed such that it will only work with `baseline` resolutions. The tests so far for tile have their config provided directly in `baseline`. However, for `pretrained` which are defined in `mpp`, a conversion and checking become necessary. However, the check is done independently from the conversion and thus the program will crash for .
https://github.com/TIA-Lab/tiatoolbox/blob/e190a0b9aa77222c8334f7eca8f3e8cd08ca285a/tiatoolbox/models/controller/semantic_segmentor.py#L598-L602
vs
https://github.com/TIA-Lab/tiatoolbox/blob/e190a0b9aa77222c8334f7eca8f3e8cd08ca285a/tiatoolbox/models/controller/semantic_segmentor.py#L255-L266

Additionally, the conversion doesnt take into account the saving resolution so it will also crash.
https://github.com/TIA-Lab/tiatoolbox/blob/e190a0b9aa77222c8334f7eca8f3e8cd08ca285a/tiatoolbox/models/controller/semantic_segmentor.py#L190-L200


tests/models/test_semantic_segmentation.py
```diff
@@ -55,6 +55,11 @@ def _rm_dir(path):
         shutil.rmtree(path, ignore_errors=True)
 
 
+def _crash_func(x):
+    """Helper to induce crash."""
+    raise ValueError("Propagation Crash.")
```
shaneahmed, line 60: Please can elaborate this error a bit more?

vqdang: Its an induced error to simulate an exception occured when the code is running. 

https://github.com/TIA-Lab/tiatoolbox/blob/30488ba15a8d7fa0caa30f0ea11b23e24de446f8/tiatoolbox/models/controller/semantic_segmentor.py#L898-L912

Depending on how we set the flag here, we either crash outright with a full stack trace printed out, or we will dump out the stack trace within a log file and carry on with the next input WSI/tile.

In either cases, it would be easier to test the above this if we can induce some specific crash pattern and exceptions. For example. I tested for crash on exception here
https://github.com/TIA-Lab/tiatoolbox/blob/948b85bf31edd009b54175e847de081884f2a865/tests/models/test_semantic_segmentation.py#L321-L332
The func above is helper to generate the pattern. Its can be of any exception type though, I only use ValueError here for conveniences.

shaneahmed (approved):
Thanks @vqdang

PR #173, opened 2021-10-27
DEV: Update links for Kather dataset

shaneahmed:
- Update links for katherdataset
- Use small sample for kather test. Full zip is only for local tests.

### DIFF
--- a/tests/models/test_dataset.py
+++ b/tests/models/test_dataset.py
@@ -92,6 +92,9 @@ def test_kather_dataset_default(tmp_path):
     ):
         _ = KatherPatchDataset(save_dir_path="unknown_place")
 
+    # remove generated data
+    shutil.rmtree(rcParam["TIATOOLBOX_HOME"])
+
 
 def test_kather_dataset(tmp_path):
     """Test for kather patch dataset."""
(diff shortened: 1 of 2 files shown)


tests/models/test_dataset.py
```diff
@@ -77,33 +77,44 @@ def __init__(self):
...
 
     # save to temporary location
     # remove previously generated data
     if os.path.exists(save_dir_path):
         shutil.rmtree(save_dir_path, ignore_errors=True)
     url = (
-        "https://zenodo.org/record/53169/files/"
-        "Kather_texture_2016_image_tiles_5000.zip"
+        "https://tiatoolbox.dcs.warwick.ac.uk/datasets"
+        "/kather100k-validation-norm-subset-90.zip"
```
shaneahmed, line 112: @John-P I think @simongraham is referring to this line.

tiatoolbox/models/dataset/info.py
```diff
@@ -83,31 +84,31 @@ def __init__(
...
         label_names = [
-            "01_TUMOR",
-            "02_STROMA",
-            "03_COMPLEX",
-            "04_LYMPHO",
-            "05_DEBRIS",
-            "06_MUCOSA",
-            "07_ADIPOSE",
-            "08_EMPTY",
+            "BACK",
```
shaneahmed, line 87: @vqdang I think this is in line with kather100k set.

PR #187, opened 2021-11-02
BUG: Fix `filter_coordinates` read wrong resolutions

vqdang:
Current predictors or any functions rely on `get_coordinates` will have the patch coordinates extracted at requested resolution not at baseline of input wsi. However, `filter_coordinates` does not indicate that and thus the `mask_reader` (after being binded to read wrt to the source image metadata) will read at wrong resolution, thus filtering out wrong coordinates.


simongraham (approved):
All looks good. Thanks for spotting the bug.

PR #202, opened 2021-11-26
ENH: Swap pygeos to shapely

vqdang:
Replace pygeos with shapely to resolve the GEO binary installation issue.

### DIFF
--- a/tests/models/test_nucleus_instance_segmentor.py
+++ b/tests/models/test_nucleus_instance_segmentor.py
@@ -20,6 +20,9 @@
 """Tests for Nucleus Instance Segmentor."""
 
 import copy
+
+# ! The garbage collector
+import gc
 import pathlib
 import shutil
 
@@ -298,6 +301,7 @@ def test_crash_segmentor(remote_sample, tmp_path):
 
 def test_functionality_travis(remote_sample, tmp_path):
     """Functionality test for nuclei instance segmentor."""
+    gc.collect()
     root_save_dir = pathlib.Path(tmp_path)
     mini_wsi_svs = pathlib.Path(remote_sample("wsi4_512_512_svs"))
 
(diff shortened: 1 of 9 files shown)


shaneahmed (approved):
Thanks @vqdang

tests/models/test_nucleus_instance_segmentor.py
```diff
@@ -319,7 +323,7 @@ def test_functionality_travis(remote_sample, tmp_path):
             {"units": "mpp", "resolution": resolution},
         ],
         margin=128,
-        tile_shape=[512, 512],
+        tile_shape=[1024, 1024],
```
John-P, line 326: Just curious why the tile size was increased. Does it help with memory? It seems like it would make it more likely to fail?

John-P (approved):
Looks good to me! Just one minor comment. I was curios about the change in tile shape.

tests/models/test_nucleus_instance_segmentor.py
vqdang: https://github.com/TissueImageAnalytics/tiatoolbox/blob/6c49bb383c1977e29833a9734dccfea7f5b26fcd/tiatoolbox/models/engine/nucleus_instance_segmentor.py#L392-L395

I have added another mode here. It is actually better because it will only run inference once for image with shape < tile shape. Tile shape was designed to handle WSI ( Height/Width > 10k). Tile shape is used to fix the tiling effect by performing re-inference along the tile boundaries.  Hence, it becomes extremely cumbersome when the tile shape is small for a small image (there will be too many unnecessary inferences. That translates to longer runtime and more accumulated patches (more failure point for OOM as by even with tile fixing, entire image must have already been inferred once already).

John-P: Thank you for clarifying, this makes a lot of sense.

PR #209, opened 2021-12-03
NEW: Easy Way To Specify / Override MPP & Power For `init` And `open`

John-P:
Suggesting a simple way to allow users to provide mpp and objective power values when opening an image / initialising a reader if they know the value.


shaneahmed (commented):
Thanks @John-P looks good to me. 
Please can you add the option to cli as well?

shaneahmed (changes requested):
This requires fixing both deepsource and travis errors.

John-P:
> Thanks @John-P looks good to me. Please can you add the option to cli as well?

How do you want to handle this? The CLI can currently take a directory of list of images. Would you like MPP and power to also be a list? Or allow a single value and a list? Throw an error if they don't match in length I assume also.

shaneahmed (approved):
Thanks @John-P

PR #210, opened 2021-12-06
BUG: Fix HoVer-Net+ nuclear post-processing

adamshephard:
- Fix a bug in the nuclear post-processing of the HoVer-Net models `HoVerNet._proc_np_hv`. The parameters used for post-processing (e.g., remove small object size, and filter kernel size) should be scaled according to the resolution of the image.


tiatoolbox/models/architecture/hovernet.py
```diff
@@ -473,6 +474,8 @@ def _proc_np_hv(np_map: np.ndarray, hv_map: np.ndarray):
             hv_map (np.ndarray): An array of shape (heigh, width, 2) which
               contains the horizontal (channel 0) and vertical (channel 1)
               of possible instances exist withint the images.
+            fx (float): The microns per pixel resolution of the input images. Default
+              is 0.5 for HoVer-Net.
```
vqdang, line 478: This is not micron per pixel, Better keep it simply as scale factor to resize kenel size or region of interests. And mention the assume size is at 0.25mpp.

tiatoolbox/models/architecture/hovernetplus.py
```diff
@@ -38,16 +38,13 @@ class HoVerNetPlus(HoVerNet):
 
     HoVer-Net+ takes an RGB input image, and provides the option to simultaneously
     segment and classify the nuclei present, aswell as semantically segment different
-    regions or layers in the images.
+    regions or layers in the images. Note HoVer-Net+ trained models are currently at
+    20X magnification, in contrast to HoVer-Net at 40X magnification.
```
vqdang, line 42: Be consistent about using power or mpp. The yaml file mentions mpp, so this should be mpp. Also, rather than the model being trained at this magnification, its the entire design of the arch and its post processing / preprocing etc.

vqdang (changes requested):
Minor changes to the docs

shaneahmed (commented):
Thanks @adamshephard I have the same comments as @vqdang

tiatoolbox/models/architecture/hovernetplus.py
adamshephard: Edited to address this.

tiatoolbox/models/architecture/hovernet.py
adamshephard: Edited to address this.

vqdang (approved):

PR #211, opened 2021-12-06
STY: Code Quality & Make Glymur Import Lazy

John-P:
- General refactor of wsireader module for code quality.
- Make glymur a lazy import (only imported when the class is used). This means that users can avoid installing openjpeg and glymur if they do not intend to use it.
- Change from using `warnings.warn` to using the logger object defined in the root module.


shaneahmed (changes requested):
Thanks @John-P  looks fine to me. please address the travis issue.

John-P:
I have split this into two PRs to help resolve test errors and am now closing this old one.

PR #217, opened 2021-12-14
ENH: Improve Mask-Based Patch Extraction Speed & Add Units Conversion

mostafajahanifar:
In the current implementation, the `filter_coordinates` function is used to filter out generated coordinates based on the input mask. This is done by reading the mask region from `mask_reader` object (which is a VirtualWSI) at the same resolution and region requested in the patch extraction. Therefore, this process can take very long specially if the whole slide mask region is big and user requests to extract small patches from highest resolution (not to mention the increase in time due to overlap in patch extraction). For example, if you want extract 256x256 patches at 40x from the below WSI, it might take up to **10 minutes** for checking filter the coordinates

output
output_

In this PR, I have introduced `filter_coordinates_fast` which uses the overview of `mask_reader` object (a numpy array) at a fixed-low resolution (mpp=8 or power=1.25) to speed of the region mask reading process and hence improve the speed of coordinate filtering and whole patch extraction processes. For example, the above image takes about **0.5 seconds** (at least **1,000x improvement in speed** in comparison to previous method).

output__

To be able to do this, I also introduce a method to `WSIReader` class that converts resolution from one unit to all other units, because the `filter_coordinates_fast` expects its input mask resolution to be in `mpp`, `power`, or `baseline` units. 

- [x] Implement fast coordinate filtering function
- [x] Implement resolution conversion function in wsireader
- [x] Writing tests
- [x] Improve coverage


tiatoolbox/utils/transforms.py
```diff
@@ -307,3 +307,74 @@ def pad_bounds(
         raise ValueError("Invalid number of padding elements.")
     signs = np.repeat([-1, 1], ndims)
     return np.add(bounds, padding * signs)
+
+
+def convert_resolution(input_res, input_unit, baseline_mpp=None, baseline_power=None):
```
John-P, line 312: Something doesn't seem right about this, but I need a while to properly go through it. Also, returning conversion to all units seems a bit like doing a lot of unnecessary work if only one conversion is being used.. Surely a conversion from one unit to another would be better.

tiatoolbox/tools/patchextraction.py
```diff
@@ -197,6 +202,73 @@ def _generate_location_df(self):
...
+            mask_resolution = [mask_resolution, mask_resolution]
+
+        tissue_mask = mask_reader.slide_thumbnail(
+            resolution=mask_resolution, units="mpp"
+        )
+
+        # checking the coordinates
+        scaled_coords = np.zeros_like(coordinates_list)
+        for i in range(4):
+            if i % 2 == 0:
```
vqdang, line 250: What is `i` ? is it index? Put comment to clarify

> assume each axis has different fx, fy

tiatoolbox/tools/patchextraction.py
```diff
@@ -197,6 +202,73 @@ def _generate_location_df(self):
...
+                scaled_coords[:, i] = np.clip(
+                    scaled_coords[:, i], 0, tissue_mask.shape[1]
+                )
+            else:
+                scaled_coords[:, i] = (
+                    coordinates_list[:, i] * coord_resolution[1] / mask_resolution[1]
+                )
+                scaled_coords[:, i] = np.clip(
+                    scaled_coords[:, i], 0, tissue_mask.shape[0]
+                )
```
vqdang, line 263: Try to refactor using numpy only

vqdang (changes requested):
Addess changes before merging. Resolution thingy needed to be dealed with by WSIMetadata. Also reading mask should be done as a different mode for `coord_space` (receiving coordinates at WSI target resolution but actual read done only at thumb resolution w/o any rescaling). These will need to be done via another PR given the urgency of finishing the paper.

tiatoolbox/tools/patchextraction.py
```diff
@@ -197,6 +202,73 @@ def _generate_location_df(self):
...
+              where N is the number of coordinate sets and K is either 2 for centroids
+              or 4 for bounding boxes. When using the default `func=None`, K should be
+              4, as we expect the `coordinates_list` to be refer to bounding boxes in
+              `[start_x, start_y, end_x, end_y]` format.
+            coord_resolution (float): the resolution at which coordinates_list are
+             generated. NOTE: resolution is expected to be in `'mpp'` units.
+
+        Returns:
+            ndarray: list of flags to indicate which coordinate is valid.
+        """
```
shaneahmed, line 229: Add new line

tiatoolbox/utils/transforms.py
```diff
@@ -307,3 +307,74 @@ def pad_bounds(
...
+    """Converts resolution between different units.
```
shaneahmed, line 313: Add new line after """

tiatoolbox/utils/transforms.py
mostafajahanifar: Thanks @John-P , I hope you find what's not right about this sooner :-D

Re returning conversions to all units: first off, most of the units are needed to be calculated during the calculation of one of them so why not keep all of them?
Also, if we allow requesting for a specific output unit, then have added an unnecessary argument to the function whereas we should still write the function like this (calculate all units) and select the desired unit from the dictionary. Otherwise, the function would comprise many `if` statement to do the right conversion!
```
if input_unit is such and requested_unit is such:
    do this calculation
```
This part should be repeated 4x4=16 times (imaging writing tests for it)! At least, I don't know a better way to write it.
Not to mention, the whole thing really need a small amount of computation (it tooks 12.2 **ns** on my laptop). So I guess we don't need to be worry about it. 
However, if you insist on returning only one desired unit, I'm happy to add another line of code the end of the function, extracting the desired unit from the output dictionary.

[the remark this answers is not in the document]
mostafajahanifar: please refer to my response to John above.

tiatoolbox/utils/transforms.py
```diff
@@ -307,3 +307,74 @@ def pad_bounds(
...
+            output_mpp = baseline_mpp / unit_baseline_scale
+        if baseline_power is not None:
+            output_power = baseline_power * unit_baseline_scale
+        output_level = np.log2(1 / unit_baseline_scale)
+
+    return {
+        "mpp": output_mpp,
+        "power": output_power,
+        "level": output_level,
+        "baseline": unit_baseline_scale,
```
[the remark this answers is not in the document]
mostafajahanifar: Default values for the parameters have been set to None. User/developer can check the values when they call the function according to their application.

tiatoolbox/utils/transforms.py
```diff
@@ -307,3 +307,92 @@ def pad_bounds(
         raise ValueError("Invalid number of padding elements.")
     signs = np.repeat([-1, 1], ndims)
     return np.add(bounds, padding * signs)
+
+
+def convert_resolution_units(
+    input_res, input_unit, baseline_mpp=None, baseline_power=None
+):
+    """Converts resolution value between different units.
```
shaneahmed, line 315: new line after """

mostafajahanifar:
Thanks @vqdang  for the review.

Re:
> Resolution thingy needed to be dealed with by WSIMetadata. 

I believe this function can be called in a WSIMeta **method**, where there is access to WSI metadata and we can control what data should be fed into the `convert_resolution_units` and check if the returned value are as expected. For example, in WSIMeta we should check if the sequence of level information stored in the WSIMeta are ordered (and raise a warning otherwise to inform the user that some processes may get wrong output).

Re:
> Also reading mask should be done as a different mode for coord_space (receiving coordinates at WSI target resolution but actual read done only at thumb resolution w/o any rescaling)

I agree with you and it will be good to have such an option in future. However, for that we need to consider scaling the `rect` parameters in the `read_rect` function for VirtuaWSI and that need further information from the user in the input about the thumb resolution, resolutions being in the same unit, avoiding extraction of thumbnail several times, etc. So, as you said, it's the future problem!

John-P (approved):

PR #222, opened 2021-12-17
ENH: On Demand Imports

John-P:
The aim of this PR (which is a branch off of tidying wsireader.py) aims to make imports only happen as needed where possible. This means that some dependencies can be avoided if they are not required.

On demand imports so far:
- glymur (and openjpeg)
- defusedxml


John-P:
Anny suggestions for additional lazy imports?

John-P:
@ghadjigeorghiou For a minimal set of requirements, I believe you could skip these out:
- jupyterlab
- glymur and openjpeg (if not using JP2)

shaneahmed (approved):
Thanks @John-P

PR #189, opened 2021-11-03
EG: Notebook For Prediction Of Molecular Pathways & Key Mutations

simongraham:
Added example notebook for mutation prediction. Include the 2 stage IDaRS pipeline to predict key mutations and molecular pathways. First, tumour detection is performed and then the WSI-level prediction task is done. WSI level prediction tasks considered:

- MSI 
- TP53
- BRAF
- CIN
- CIMP
- HM

As part of this PR, we retrained IDaRS models without stain normalisation to speed up inference.

PR #225, opened 2021-12-17
NEW: Add PCam Patch Classification Models

simongraham:
Add patch classification models trained on the [Patch Camelyon dataset](https://github.com/basveeling/pcam). We include all models that were used for the Kather100K dataset. 

As part of this PR, we : 
- Add new model links
- Add tests for model output
- Add links to 2 new sample patches for testing


shaneahmed (commented):
looks fine to me except idars notebook. 
is it not part of #189 ?

PR #229, opened 2021-12-21
API: Use normalize (US Spellings) instead of normalise (UK Spellings)

shaneahmed:
- Use normalize (US Spellings) instead of normalise (UK Spellings)


simongraham (changes requested):
All looks good - however, we just need to change the name of the stain norm example notebook

PR #258, opened 2022-01-17
MAINT: Setting Up Conditional Tests Using "pytest skipif"

vqdang:
There are several tests that have been marked to be skipped when running pytest because there were no way to check if they are run on Travis. Now we have the facility for detecting runtime flag in place, we can automatically flag which test to run on which conditions. This PR deals with tests that are related to Deep Learning Models (require GPU and Travis runtime check).

### DIFF
--- a/tests/models/test_semantic_segmentation.py
+++ b/tests/models/test_semantic_segmentation.py
@@ -677,7 +677,7 @@ def test_functional_pretrained(remote_sample, tmp_path):
 
 
 @pytest.mark.skipif(
-    toolbox_env.running_on_travis() or not toolbox_env.has_gpu(),
+    toolbox_env.running_on_travis() or not ON_GPU,
     reason="Local test on machine with GPU.",
 )
 def test_behavior_tissue_mask_local(remote_sample, tmp_path):
@@ -718,7 +718,7 @@ def test_behavior_tissue_mask_local(remote_sample, tmp_path):
 
 
 @pytest.mark.skipif(
-    toolbox_env.running_on_travis() or not toolbox_env.has_gpu(),
+    toolbox_env.running_on_travis() or not ON_GPU,
     reason="Local test on machine with GPU.",
 )
 def test_behavior_bcss_local(remote_sample, tmp_path):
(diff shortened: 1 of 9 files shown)


tests/models/test_semantic_segmentation.py
```diff
@@ -44,11 +45,15 @@
     SemanticSegmentor,
     WSIStreamDataset,
 )
+from tiatoolbox.utils import env_detection as toolbox_env
 from tiatoolbox.utils.misc import imread, imwrite
 from tiatoolbox.wsicore.wsireader import get_wsireader
 
-ON_TRAVIS = True
-ON_GPU = not ON_TRAVIS and torch.cuda.is_available()
+ON_GPU = toolbox_env.has_gpu()
```
shaneahmed, line 52: this variable is not used very much. I would prefer to skip it or use it everywhere.

tests/test_utils.py
```diff
@@ -1248,3 +1248,11 @@ def test_detect_travis():
     Simply check it passes without exception.
     """
     _ = utils.env_detection.running_on_travis()
+
+
+def test_detect_gpu():
+    """Test detection of GPU in the current runtime environment.
+
+    Simply check it passes without exception.
+    """
```
shaneahmed, line 1257: add a new line at the end of docstring.

shaneahmed (commented):
@vqdang  this PR looks good. Please can you address the comments? I can then approve.

tests/models/test_semantic_segmentation.py
vqdang: I dont understand. `ON_GPU` is used at 14 different places in the files.

shaneahmed: you are still using `toolbox_env.has_gpu()` within the file. I am suggesting to use the variable everywhere or just not use it at all. just to be consistent.

vqdang: Fixed in https://github.com/TissueImageAnalytics/tiatoolbox/pull/258/commits/9260d9c2d654ac6f097cce2e8468e9c24fccb7ad

shaneahmed (approved):
Thanks @vqdang 
Let's get another review from Simon and merge this.

simongraham (approved):

PR #261, opened 2022-01-17
BUG: Add User Warning For Missing SQLite Functions

John-P:
It is possible that SQLite (e.g. versions <3.35) may not have math functions enabled (e.g. floor). This could break some queries e.g. those made using a DSL string including a floor division (//).

Added information to docs. See https://tia-toolbox.readthedocs.io/en/bug-sqlite-floor/_autosummary/tiatoolbox.annotation.dsl.html.

### DIFF
--- a/tiatoolbox/annotation/storage.py
+++ b/tiatoolbox/annotation/storage.py
@@ -77,6 +77,7 @@ from shapely.geometry import LineString, Point, Polygon
 from shapely.geometry import mapping as geometry2feature
 from shapely.geometry import shape as feature2geometry
 
+from tiatoolbox import logger
 from tiatoolbox.annotation.dsl import (
     PY_GLOBALS,
     SQL_GLOBALS,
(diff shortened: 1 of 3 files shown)


tiatoolbox/annotation/storage.py
```diff
@@ -1003,6 +1004,14 @@ def __init__(
...
             raise Exception("RTREE and JSON1 sqlite3 compile options are required.")
 
+        # Check that math functions are enabled
+        if "ENABLE_MATH_FUNCTIONS" not in compile_options:
+            logger.warning(
+                "SQLite math functions are not enabled."
+                " This may cause problems with some queries."
+                " For example, floor division (//) will not work."
+            )
+
```
vqdang, line 1014: It would be more helpful if you put the cases that could use // underneath within the doctring, and then add the URL to readthedoc in the warning here for fixing direction.

vqdang (commented):
Minor comment.

tiatoolbox/annotation/storage.py
John-P: Just added a documentation build to check this. Currently I think it is only `//` that will be affected.

John-P: See https://tia-toolbox.readthedocs.io/en/bug-sqlite-floor/_autosummary/tiatoolbox.annotation.dsl.html

vqdang: Looking good

vqdang (approved):

ghadjigeorghiou (approved):

PR #273, opened 2022-01-31
ENH: Fix Bugs & Improve UNet Architecture

mostafajahanifar:
There are two bugs in the current implementation of UNet architecture **when UNet encoder is selected:**
1- Number of channels in Batchnorm argument in the decoding path does not match with the input channels.
2- Padding 0 does not create feature maps in the decoder part with the same size as encoder. Also, fixed padding won't work when user selects kernel sizes other than 3. 

Therefore, I have made the following changes to fix these issues and improve the architecture definition:

- [x] Fix batchnorm bug
- [x] Fix padding. Use `padding='same'` instead of padding=1 in both unet and resnet cases to avoid problems when `kernel_size!=3` is used.
- [x] Expose encoder configuration in the UNetModel definition as `encoder_levels` argument.
- [x] Add the support for both "add" and "concat" skip connections
- [x] Fix flake-8 errors
- [x] Avoid preprocessing in "forward" function (it can be problematic, also, it reduces flexibility in model use).
- [x] Update unit tests.


tiatoolbox/models/architecture/unet.py
```diff
@@ -220,20 +226,33 @@ def __init__(
         num_input_channels: int = 2,
         num_output_channels: int = 2,
         encoder: str = "resnet50",
-        decoder_block: Tuple[int] = (3, 3),
+        encoder_levels: List[int] = None,
+        decoder_block: Tuple[int] = None,
+        skip_type: str = "add",
```
vqdang, line 231: Intial design was to hard lock only 2 form of encoders. Now it becomes quite counter intuitive between `encoder_levels` and `encoder`. Should retain hard lock and expose overide for `encoder` (as in receiving a model).

tiatoolbox/models/architecture/unet.py
```diff
@@ -220,20 +226,33 @@ def __init__(
...
+
         if encoder == "resnet50":
-            padding = 1
+            padding = "same"
             preact = True
             self.backbone = ResNetEncoder.resnet50(num_input_channels)
         elif encoder == "unet":
-            padding = 0
+            padding = "same"
             preact = False
```
vqdang, line 250: Back to above about hard locking decision. Should fix the padding below within the decoder creation rather than using "same". Additionally, do not use "same" or "valid" to maintain backward compatibility (only supported 1.9.0 onward)
https://pytorch.org/docs/1.8.0/_modules/torch/nn/modules/conv.html#Conv2d

vqdang: Fixing within `create_block`

tiatoolbox/models/architecture/unet.py
```diff
@@ -293,6 +312,8 @@ def create_block(kernels, input_ch, output_ch):
             next_up_ch = ch
             if ch_idx + 2 < len(down_ch_list):
                 next_up_ch = down_ch_list[ch_idx + 2]
+            if self.skip_type == "concat":
+                ch *= 2
```
vqdang, line 316: Ill condition, if you decide to allow arbitray `encoder_levels`, the number of channels are no longer guarantee to be a multiplication of 2 but of arbitrary number. A dummy forward within is a must to retrieve level and set the decoder input.

tiatoolbox/models/architecture/unet.py
```diff
@@ -329,10 +350,12 @@ def forward(self, imgs: torch.Tensor, *args, **kwargs):
             # coming from the encoder, then run it through the decoder
             # block
             y = en_list[-idx]
-            x = self.upsample2x(x) + y
+            if self.skip_type == "add":
+                x = self.upsample2x(x) + y
+            else:
+                x = torch.cat([self.upsample2x(x), y], dim=1)
```
vqdang, line 356: Refactor `self.upsample2x(x)` out

vqdang (changes requested):
Major revisions. Please clarify what do you mean by

> Avoid preprocessing in "forward" function (it can be problematic, also, it reduces flexibility in model use).

In case you mean
https://github.com/TissueImageAnalytics/tiatoolbox/blob/9a3bb725e0806d6d483afc93b6989048d1d66dde/tiatoolbox/models/architecture/unet.py#L319

Note that inception from pytorch is the same, and in general, offloading as much as possible to GPU is preferred. This can  also be considered as network design choice, such as HoVerNet (hard lock condition).

https://github.com/pytorch/vision/blob/8aad0e0a64c65b4e16855764288f0bc2ad2e14c0/torchvision/models/inception.py#L165

vqdang:
Please also verify that the new model is backward compatible. (Loading model weights with strict mode).

tiatoolbox/models/architecture/unet.py
mostafajahanifar: encoder_levels has been incorporated to support for the flexibility of Unet encoder choice. This is only applicable for UNet encoder as described in the docstrings

tiatoolbox/models/architecture/unet.py
mostafajahanifar: As we discussed before,  the number of channel don't need to be multiplication of 2 but they will be multiplied by 2. With the genius design of encoder-decoder parts we can make sure that number of feature map from skip connection are the same as the number feature maps from upsampling part in the decoder. I have checked the code and resulted network architecture, this works with both custom number of layers and feature maps.

tiatoolbox/models/architecture/unet.py
mostafajahanifar: done in 7da8262

tiatoolbox/models/architecture/unet.py
mostafajahanifar: done in 7da8262

mostafajahanifar:
> Please also verify that the new model is backward compatible. (Loading model weights with strict mode).

Thanks Dang for the thorough review. All your comments have been incorporated. Also, backward compatibility has been checked both using pytest and running example notebooks.

tiatoolbox/models/architecture/unet.py
```diff
@@ -1,10 +1,10 @@
-"""Defines a set of ResNet variants to be used within tiatoolbox."""
+"""Defines a set of UNet variants to be used within tiatoolbox."""
 
 from typing import List, Tuple
 
 import torch
 import torch.nn as nn
-import torch.nn.functional as F
+import torch.nn.functional as nn_func
```
shaneahmed, line 7:
```suggestion
import torch.nn.functional as functional
```

shaneahmed:
@mostafajahanifar Please can you have a look at this https://github.com/TissueImageAnalytics/tiatoolbox/pull/345

tiatoolbox/models/architecture/unet.py
mostafajahanifar: Do you want me to fix it in this PR or #345?

shaneahmed: Let's fix this in this PR. I can close the other PR once we have merged this PR.

BTW, what's the status of this PR? Do you need reviews on this PR still?

mostafajahanifar: The PR is ready to be merged really, it has been reviewed several time and applied all the comments. However, I requested for @simongraham  and @vqdang but not received any feedback yet.

shaneahmed: @simongraham and @vqdang  please can you confirm if you are happy with this PR?

tiatoolbox/models/architecture/unet.py
```diff
@@ -1,10 +1,10 @@
-"""Defines a set of ResNet variants to be used within tiatoolbox."""
+"""Defines a set of UNet variants to be used within tiatoolbox."""
 
 from typing import List, Tuple
 
 import torch
 import torch.nn as nn
-import torch.nn.functional as F
+import torch.nn.functional as functional
```
vqdang, line 7: Not sure why this needed to be changed, `F` is the common form for pytorch code so it should be best to keep it in line with the community. You do no not need to edit but should keep this in mind.

vqdang (approved):
I am ok, but will put here as a comment and warning (?)

`import torch.nn.functional as F` is the common form for pytorch code so it should be best to keep it in line with the community. You do no not need to edit but should keep this in mind.

tiatoolbox/models/architecture/unet.py
John-P: We discussed this in the last toolbox meeting. I agree that sticking to what the community does would make sense so that when people read our code, it matches up with other common code e.g.:

```python
import torch.nn.functional as F
from torch import nn
from matplotlib import pyplot as plt
# etc.
```

PR #283, opened 2022-02-08
BUG: Fix Graph Functions

John-P:
This PR fixes some issues in graph.py

1. Fix typos in docstrings.
2. Fix typo in function name traingle -> triangle
3. Fix a bug in graph visualisation where some edges would not be drawn.

Some other things have been fixed as they were discovered during this PR:
1. Fix a bug in function to convert edge index to triangles (it would find the same triangle twice in a different order).
2. Add tests.

### DIFF
--- a/tiatoolbox/tools/graph.py
+++ b/tiatoolbox/tools/graph.py
@@ -217,7 +217,7 @@ class SlideGraphConstructor:  # noqa: PIE798
         Args:
             graph (dict):
                 A graph with keys "x", "edge_index", and optionally
-                "coords".
+                "coordinates".
         Returns:
             ArrayLike: A UMAP embedding of `graph["x"]` with shape (N, 3)
                 and values ranging from 0 to 1.
@@ -396,7 +396,7 @@ class SlideGraphConstructor:  # noqa: PIE798
         return {
             "x": feature_centroids,
             "edge_index": edge_index,
-            "coords": point_centroids,
+            "coordinates": point_centroids,
         }
 
     @classmethod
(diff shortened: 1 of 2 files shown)


tiatoolbox/tools/graph.py
```diff
@@ -283,7 +283,7 @@ def build(
             feature_range_thresh (Number):
                 Minimal range for which a feature is considered significant.
                 Features which have a range less than this are ignored.
-                Defaults to 1e-4. If falsey (None, False, 0, etc.), then
+                Defaults to 1e-4. If falsy (None, False, 0, etc.), then
```
shaneahmed, line 286:
```suggestion
                Defaults to 1e-4. If False (None, False, 0, etc.), then
```

shaneahmed (commented):
just a minor comment

tiatoolbox/tools/graph.py
John-P: I specifically say falsy not False here because `0` and `None` are not false, but they are 'falsy' i.e. convert to `False` when `__bool__` is used. E.g. `bool(None) == False` even though `None != False`. Cna change the wording if you like. Truthy and falsy are commonly used to describe this behaviour, although they are more commonly used in javascript. See https://docs.python.org/3/library/stdtypes.html#truth and https://developer.mozilla.org/en-US/docs/Glossary/Falsy for more.

shaneahmed (approved):

vqdang (approved):
Looks fine to me

vqdang:
Brief check on notebook to make sure

PR #284, opened 2022-02-11
DEV: Update scikit-image Version

Srijay-lab:
- Update scikit image version to 0.19.1

### DIFF
--- a/requirements.win64.conda.yml
+++ b/requirements.win64.conda.yml
@@ -31,7 +31,7 @@ dependencies:
   - pytorch>=1.9.0
   - pyyaml>=5.1
   - requests
-  - scikit-image==0.19.1
+  - scikit-image>=0.19.1
   - scikit-learn>=0.23.2
   - shapely
   - sphinx-gallery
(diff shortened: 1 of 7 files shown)


requirements.win64.conda.yml
```diff
@@ -31,7 +31,7 @@ dependencies:
   - pytorch>=1.9.0
   - pyyaml>=5.1
   - requests
-  - scikit-image==0.18.3
+  - scikit-image==0.19.1
```
shaneahmed, line 34:
```suggestion
  - scikit-image>=0.19.1
```

We need to test with `>=19.1`

John-P:
Could we add a comment next to this line to explain why this has to be pinned. This way, when we come back in future, we will know why it is piined.

John-P:
We should update all of the requirements files if possible to keep the versions consistent.

John-P:
Currently investigating why the scikit-image update changes the `phase_cross_correlation` and breaks tests.

shaneahmed:
@Srijay-lab  please also update `setup.py`

shaneahmed:
Please check this https://scikit-image.org/docs/stable/release_notes.html#api-changes

shaneahmed (approved):
Thanks @Srijay-lab

PR #285, opened 2022-02-14
BUG: Fixed performance issue in semantic_segmentor.py

ByteHexler:
**This PR fixes a major performance issue in [semantic_segmentor.py](https://github.com/TissueImageAnalytics/tiatoolbox/pull/285/commits/e4a64386e9c95cb5141f56044a0748621c60f89d#diff-81374c91b9cee6bb401e9f511e940b670d5a9aac2cac6795894e91d7c0be13a0)**

Moved function self._precess_predictions() out of for loop, which was causing major (unnecessary) performance loss for segmentation prediction - especially on larger images/wsi => now up to 12x faster

### DIFF
--- a/tiatoolbox/models/engine/semantic_segmentor.py
+++ b/tiatoolbox/models/engine/semantic_segmentor.py
@@ -668,11 +668,13 @@ class SemanticSegmentor:
 
             sample_outputs = list(zip(sample_infos, sample_outputs))
             cum_output.extend(sample_outputs)
-            # TODO: detach or hook this into a parallel process
-            self._process_predictions(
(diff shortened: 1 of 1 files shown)


shaneahmed:
This is deliberately done to avoid crashing the code on systems with low memory.
Please create an issue and we will work on a solution.
@vqdang do you anything to add?

shaneahmed:
> As "cum_output" is extended and then processed in each execution I figured that the last for loop execution would do the same as just having "self._precess_predictions()" after the loop. I am going to create an issue nonetheless.

Thanks, we will look into this in the meanwhile.

vqdang:
Migrated to #313 for dedicated tracking/fixing PR from the internal team.

PR #292, opened 2022-02-18
BUG: Fix Travis Detection 

shaneahmed:
- tox.ini modified to pass Travis variables which fixes the tests.
- Updated tests to run locally for Travis detection.
- Reduces Travis runtime by up to 5 mins per run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ install:
     - source "$HOME/miniconda/etc/profile.d/conda.sh"
     - pip install -U tox-travis
 
+
 # Command to run tests, e.g. python setup.py test
 script:
     - set -e # Exit on first failure (errexit)
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,9 @@ install:
     - source "$HOME/miniconda/etc/profile.d/conda.sh"
     - pip install -U tox-travis
 
+env:
+  global:
+    - ON_TRAVIS=true
 
 # Command to run tests, e.g. python setup.py test
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,6 @@ install:
     - source "$HOME/miniconda/etc/profile.d/conda.sh"
     - pip install -U tox-travis
 
-env:
-  global:
-    - ON_TRAVIS=true
-
 # Command to run tests, e.g. python setup.py test
 script:
     - set -e # Exit on first failure (errexit)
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -20,6 +20,9 @@ install:
     - source "$HOME/miniconda/etc/profile.d/conda.sh"
     - pip install -U tox-travis
 
+env:
+  global:
+    - ON_TRAVIS=true
```
John-P, line 26: Why are you defining a new env variable here? Travis already defines two `TRAVIS` and `CI`. Why not just use those?

tests/test_utils.py
```diff
@@ -1255,7 +1255,11 @@ def test_detect_travis():
 
     Simply check it passes without exception.
     """
-    _ = utils.env_detection.running_on_travis()
+    on_travis = utils.env_detection.running_on_travis()
+    if "ON_TRAVIS" in os.environ:
+        assert on_travis
+    else:
+        assert not on_travis
```
John-P, line 1262: Should be using `TRAVIS` instead of `ON_TRAVIS` here

.travis.yml
John-P: See  https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

tests/test_utils.py
John-P: See https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

John-P, line 1262: Not sure about this. I think it was better before simply checking that it returns. Maybe could add a check that it is a boolean value. This looks like it will easily break as a test and isn't actually testing anything,

shaneahmed: `utils.env_detection.running_on_travis()` is dependent on Travis variables. So I have forced a test which is relying on a variable which is manually added in case the other variables fail.
This will allow the test to pass locally as well otherwise tests will fail locally.

.travis.yml
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/292/commits/a16c0188a644d52be66443f445fd552f067e5341

tests/test_utils.py
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/292/commits/22768389760785e09e96e83bb7d1a33e6fa2c2e7

PR #293, opened 2022-02-18
NEW: Add `micronet`

shaneahmed:
- Add `micronet` architecture.
- Add Consep trained model.


tiatoolbox/models/architecture/micronet.py
```diff
@@ -0,0 +1,567 @@
...
+            input_tensor (torch.Tensor): Input images, the tensor is
+                in the shape of NCHW.
+
+        Returns:
+            list: A list of main and auxiliary outputs.
+                The expected format is [main_output, aux1, aux2, aux3].
+
+        """
+
+        def group1_branch(layer, in_tensor, resized_feat):
```
adamshephard, line 373: Would the below group functions be better defined in __init__ under a slight different name such as group_link? Makes forward quite clunky otherwise.

shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/293/commits/9af1c008012edf59b1599184b900515b412365d0

vqdang (changes requested):
@shaneahmed Can you add a functional test for the precision of the output? You will use the conditional decorator to flag that test for local run only?

Otherwise, I am fine with this.

vqdang (approved):

PR #307, opened 2022-03-03
DEV: Improve CLI

shaneahmed:
- Restructure cli to use lazy imports
- Minimize duplication for consistency in input arguments using common decorators.


tiatoolbox/cli/common.py
```diff
@@ -0,0 +1,166 @@
...
+        super(TIAToolboxCLI, self).__init__(*args, **kwargs)
+        self.help = "Computational pathology toolbox by TIA Centre."
+        self.add_help_option = {"help_option_names": ["-h", "--help"]}
+
+
+def no_input_message(message):
+    """This function is called if no input is provided.
+
+    Args:
+        message (str): Failure message to display.
```
shaneahmed, line 66:
```suggestion
        message (str): Error message to display.
```

tiatoolbox/cli/slide_info.py
```diff
@@ -20,54 +20,57 @@
...
-)
-@click.option(
-    "--verbose",
-    type=bool,
-    default=True,
-    help="Print output, default=True",
-)
+@slide_info_cli.command()
+@cli_img_input
+@cli_output_path
```
shaneahmed, line 41: @John-P  Is there a way to edit the "help" part or default values for this decorator?

John-P: Can I just check exactly which decorator you mean? There are many here

John-P: If so, then simply make the function in `common` into a decorator itself which has a default help argument which you could then override.

shaneahmed: Please can you provide an example to explain?

shaneahmed: Actually, got it just trying this. I will let you know if I need help on this.

tests/test_slide_thumbnail.py
```diff
@@ -68,9 +68,7 @@ def test_command_line_slide_thumbnail_output_none(sample_svs, tmp_path):
 
     assert slide_thumb_result.exit_code == 0
     assert (
-        pathlib.Path(sample_svs).parent
-        / ("slide-thumbnail")
-        / (sample_svs.stem + ".jpg")
+        pathlib.Path(sample_svs).parent / ("output") / (sample_svs.stem + ".jpg")
```
John-P, line 71: Output doesn't need brackets here. It doesn't matter, but just looks a little odd.

tiatoolbox/cli/slide_info.py
John-P: If you want to wrap the decorator, you can do something like this:
```python
def cli_output_path(help:str="Path to output directory to save the output.", *args, **kwargs):
    inner_func = click.option(
        "--output-path",
        help=help,
    )
    return inner_func 
```

Might not be precisely right but that is the gist of it.

tiatoolbox/cli/slide_info.py
```diff
@@ -20,54 +20,57 @@
...
+    no_input_message,
+    prepare_file_dir_cli,
+)
 
-
-@click.group()
-def main():  # pragma: no cover
-    """Define slide_info click group."""
-    return 0
+slide_info_cli = TIAToolboxCLI()
```
John-P, line 36: Looks like you create this object in each file. Is that the intended way from click? I am not sure because I do not know the library that well. It looks like you want this to be a singleton which inheriting from Group might make it but I am not sure.

John-P (approved):
Looks OK to me. I am not super familiar with click, so I can't give feedback specific if this is the 'right' way to do it. But if it works, then I am happy. Certainly looks a bit tidier.

tiatoolbox/cli/slide_info.py
shaneahmed: I am creating these for now but will address this in a separate PR.
The plan is to create an abstract class with subclasses for each command which invokes only if called with lazy imports.

tiatoolbox/cli/slide_info.py
shaneahmed: Resolved in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/307/commits/585f88d086b8ebc6a1f65b22a7a59c30dd3ec35d

tiatoolbox/cli/slide_info.py
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/307/commits/c19e3cc7d4ac22cd26b37ca40b0790edf71d27ce

tests/test_slide_thumbnail.py
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/307/commits/c19e3cc7d4ac22cd26b37ca40b0790edf71d27ce

tests/test_wsireader.py
```diff
@@ -1642,6 +1642,10 @@ def test_command_line_jp2_read_bounds(sample_jp2, tmp_path):
     assert pathlib.Path(tmp_path).joinpath("../im_region.jpg").is_file()
 
 
+@pytest.mark.skipif(
+    utils.env_detection.running_on_travis(),
+    reason="No need to display image on travis.",
+)
```
John-P, line 1648: Just wondering why this is added? Is it slow on travis?

John-P (approved):

tests/test_wsireader.py
shaneahmed: Just avoiding the clunky output on travis.

PR #316, opened 2022-03-11
DEV: Fix Makefile Docs Command

John-P:
There was a bug in `make docs` where it would try to remove a directory `rm docs/_autosummary` without the required `-r` option. This has been fixed in this PR. Also,  a `clean-docs` command has been added in addition to including generated files in the .gitignore file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,10 +49,12 @@ clean-test: ## remove test and coverage artifacts
 
 clean-docs: ## remove documentation artifacts
 	rm -fr docs/_build/
+	rm -rf docs/_autosummary/
+	rm -rf docs/_notebooks/
 	rm -fr docs/html/
 	rm -f docs/tiatoolbox.rst
 	rm -f docs/modules.rst
-	rm -rf docs/_autosummary/
+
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -47,6 +47,15 @@ clean-test: ## remove test and coverage artifacts
 	rm -fr htmlcov/
 	rm -fr .pytest_cache
 
+clean-docs: ## remove documentation artifacts
+	rm -fr docs/_build/
```
shaneahmed, line 51: what's the difference between `-fr` and `-rf`? If both are same why not just use `-rf`

shaneahmed (approved):
Please can you address my comment? Otherwise it looks fine to me.

Makefile
John-P: They are the same, -f meant 'force' and -r means 'recursive' -rf and -fr means both.

vqdang:
Have you tested running `make doc` under root directory rather than `./doc`? `readthedoc` only cares about the make file in the `doc` if i rmb it right.

Issue #314, opened 2022-03-08
get_wsireader doesn't work for .tiff files converted from jp2


R-J96:
* TIA Toolbox version: 1.0.1
* Python version: 3.7.8
* Operating System: Ubuntu 20.04

### Description

Using `get_wsireader` on a tiff files converted from jp2 errors but using the OpenSlideWSIReader class works. This in turn means classes like SemanticSegmentor that use `get_wsireader` also don't work with this.

### What I Did
Code to reproduce this
```
wsi = get_wsireader(".../file.tiff")
```
results in following error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/robj/.conda/envs/tiatoolbox/lib/python3.7/site-packages/tiatoolbox/wsicore/wsireader.py", line 2417, in get_wsireader
    return WSIReader.open(input_img)
  File "/home/robj/.conda/envs/tiatoolbox/lib/python3.7/site-packages/tiatoolbox/wsicore/wsireader.py", line 134, in open
    return VirtualWSIReader(input_img, mpp=mpp, power=power)
  File "/home/robj/.conda/envs/tiatoolbox/lib/python3.7/site-packages/tiatoolbox/wsicore/wsireader.py", line 1816, in __init__
    self.img = utils.misc.imread(self.input_path)
  File "/home/robj/.conda/envs/tiatoolbox/lib/python3.7/site-packages/tiatoolbox/utils/misc.py", line 160, in imread
    image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
cv2.error: OpenCV(4.5.5) /io/opencv/modules/imgproc/src/color.cpp:182: error: (-215:Assertion failed) !_src.empty() in function 'cvtColor'
```

but using
```
wsi = OpenSlideWSIReader(".../file.tiff")
```
on the same file runs absolutely fine.

TIFFWSIReader gives a different error
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/robj/.conda/envs/tiatoolbox/lib/python3.7/site-packages/tiatoolbox/wsicore/wsireader.py", line 2053, in __init__
    raise ValueError("Unsupported TIFF WSI format.")
ValueError: Unsupported TIFF WSI format.
```

shaneahmed:
The issue is being fixed in #317 
Just a comment, please use WSIReader.open() as get_wsireader will be depreciated in future release.

PR #317, opened 2022-03-11
ENH: Add Better Support For Tiled TIFFs

John-P:
- Add a check for tiles in a TIFF file when opening 
- Use OpenSlide to read a TIFF if it has tiles instead of OpenCV (VirtualWSIReader).
- Add a fallback to tifffile if it is tiled but openslide cannot read it (e.g. jp2k or jpegxl tiles).


John-P:
This is currently failing due to now supporting more TIFF formats, and we don't currently have an example for which this would fail. Therefore, either the test coverage will be reduced which will fail the PR or the test will fail which will fail the PR. I will look into generating a test image where this should fail.

John-P:
I have made a simple test image which is a copy of CMU-1-Small-Region.svs but with two tiled pages and no resolution metadata. This is used as an example fail case, as we currently don't support multipage generic tiled TIFF. Only known TIFF formats such as SVS, NDPI, OME-TIFF, and generic single-page tiled TIFFs. This could be extended to single series, e.g. one series of pyramid resolutions in future. This test image would also fail that, as there is no subifd tag for reduced resolution set.

tiatoolbox/utils/misc.py
```diff
@@ -756,3 +756,30 @@ def string_to_tuple(in_str):
...
+    """
+    microns_per_unit = {
+        "centimeter": 1e4,  # 10,000
+        "cm": 1e4,  # 10,000
+        "inch": 25400,
+        2: 25400,  # inches in TIFF tags
+        3: 1e4,  # cm in TIFF tags
+    }
+    if units not in microns_per_unit:
+        raise ValueError(f"Invalid units: {units}")
```
shaneahmed, line 784: A small test can be added with invalid units to raise the error.

shaneahmed (commented):
Thanks @John-P 
Please can you add the test as mentioned?

shaneahmed:
@R-J96 As this is related to #314. Please can you review this PR?

R-J96 (approved):

shaneahmed (approved):

John-P:
@shaneahmed My current solution to the deepsource path issue is a combination of `# skipcq: PTC-W6004` it ignore generated paths in the whole file:
https://github.com/TissueImageAnalytics/tiatoolbox/blob/44c59b443bf4381892f7bb0487d48bbebc580707/tests/test_annotation_stores.py#L1

and allowing commented out code in tests (vecause flake thinks that skipcq is code!):
https://github.com/TissueImageAnalytics/tiatoolbox/blob/44c59b443bf4381892f7bb0487d48bbebc580707/setup.cfg#L25-L26

PR #319, opened 2022-03-18
TST: Make Stain Norm CLI Tests Use Temp Dir

John-P:
Fix the stain normalisation CLI test which would dump output into the project root directory by specifying an output to a temporary test directory.


shaneahmed (approved):
Thanks @John-P  Looks fine to me.

PR #321, opened 2022-03-24
NEW: CLI For TileServer

measty:
Changed this commit to just add a CLI for setting up a tileserver


tiatoolbox/cli/overlay_from_json.py
```diff
@@ -0,0 +1,169 @@
...
+        "uint32": "uint",
+        "int32": "int",
+        "float32": "float",
+        "float64": "double",
+        "complex64": "complex",
+        "complex128": "dpcomplex",
+    }
+    height, width, bands = a.shape
+    linear = a.reshape(width * height * bands)
+    vi = pyvips.Image.new_from_memory(
```
shaneahmed, line 30: Can we not avoid pyvips dependancy? 
Adding it would mean further restrictions on dependency versions.

John-P: Indeed tifffile / zarr cna be used instead here.

John-P:
I like this idea. However, my main concern is with the code structure. It doesn't make sense to have the whole implementation in a cli function. I would implement rendering if tiles from an annotation store first. Then a function to write all the tiles to a tiff with tiffile/zarr. Then make a wrapper cli argument which loads the geojson (or ndgeojson) to a store, then renders is out.

tests/test_tileserver.py
```diff
@@ -1,35 +1,79 @@
...
+        cli.main,
+        [
+            "slide-thumbnail",
+            "--img-input",
+            str(Path(sample_ndpi)),
+            "--mode",
+            "save",
+        ],
+    )
+    wsi = WSIReader.open(Path(sample_ndpi))
```
shaneahmed, line 29: why do you need this?

tiatoolbox/cli/tile_server.py
```diff
@@ -0,0 +1,17 @@
...
+)
+
+
+@tiatoolbox_cli.command()
+@click.argument("img_paths", nargs=-1, type=click.Path(exists=True))
+def tile_server(img_paths):  # pragma: no cover
+    """show a slide together with a whole slide overlay"""
+    from tiatoolbox.visualization.tileserver import TileServer
+
+    app = TileServer("TileServer", list(img_paths))
```
shaneahmed, line 15: Can we not run the server in a directory and open image directly from there?

shaneahmed (commented):
Thanks Mark. Please can you address the comments.

tests/test_tileserver.py
measty: Hi Shan,
I use it in line 40: `np.zeros(wsi.slide_dimensions(1.25, "power"), dtype=np.uint8).T` to make an array of suitable dimensions to test the case where a low-res heatmap is given as an np array for one of the layers

tiatoolbox/cli/tile_server.py
measty: do you mean just give a directory and try to open all images in there as layers?

shaneahmed: We can open individual images here, but yes we should be able to open a directory in future.

shaneahmed: I am resolving it for the sake of this PR.

shaneahmed (approved):
Thanks @measty 
I will wait for @John-P's input before I merge it.

tiatoolbox/cli/tile_server.py
```diff
@@ -0,0 +1,17 @@
+"""Command line interface for TileServer"""
+import click
+
+from tiatoolbox.cli.common import (
+    tiatoolbox_cli,
+)
+
+
+@tiatoolbox_cli.command()
+@click.argument("img_paths", nargs=-1, type=click.Path(exists=True))
```
shaneahmed, line 10: shouldn't this be using decorator from `cli/common.py` for image input? This will automatically add help for that function as well. https://github.com/TissueImageAnalytics/tiatoolbox/blob/446f2e1e05f79d4e397267a2dea4f962c506fdd1/tiatoolbox/cli/common.py#L30

measty: the img_input decorator will only allow a single path (to a file or directory). But the tileserver cli needs to be able to accept a variable number of paths.
for example if i had a slide and heatmaps from two different models i need to be able to do:

tiatoolbox tileserver path/to/slide path/to/heatmap1 path/to/heatmap2

to start the tileserver with the slide and two additional layers for the heatmaps

shaneahmed: You can still use `img_input` by defining defaults and options. See for example, https://github.com/TissueImageAnalytics/tiatoolbox/blob/c7fe7851cf78ab582e8e02bc1b21d5227063fc77/tiatoolbox/cli/common.py#L50 You can see different file types have been called here https://github.com/TissueImageAnalytics/tiatoolbox/blob/c7fe7851cf78ab582e8e02bc1b21d5227063fc77/tiatoolbox/cli/patch_predictor.py#L32 
Similarly, options for `mode` have been modified https://github.com/TissueImageAnalytics/tiatoolbox/blob/c7fe7851cf78ab582e8e02bc1b21d5227063fc77/tiatoolbox/cli/patch_predictor.py#L35

measty: the only options available for img_input is the help option. 

https://github.com/TissueImageAnalytics/tiatoolbox/blob/c7fe7851cf78ab582e8e02bc1b21d5227063fc77/tiatoolbox/cli/common.py#L30-L32

I could change that to allow more options to cover the useage needed here, if that would be a better way to do it?

shaneahmed: yes that's fine you can modify it by keeping the previous options as default options.

tiatoolbox/visualization/tileserver.py
```diff
@@ -37,7 +39,7 @@ class TileServer(Flask):
         >>> app.run()
     """
 
-    def __init__(self, title: str, layers: Dict[str, WSIReader]) -> None:
+    def __init__(self, title: str, layers: Union[Dict, List]) -> None:
```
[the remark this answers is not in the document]
measty: >I did this so that it isn't necessary to give names to the layers. You can give it a list and have autogenerated layer names ('layer1', 'layer2', etc), or a dict which specifies names for each layer via the key.

Although, thinking on that, there isn't a provision for specifying the layer names in the cli command, and there probably should be. I can add that in later today.

Will change the hint as suggested also.

tiatoolbox/cli/overlay_from_json.py
John-P: I think this is resolved now.

tests/test_tileserver.py
```diff
@@ -1,35 +1,79 @@
...
+    runner.invoke(
+        cli.main,
+        [
+            "slide-thumbnail",
+            "--img-input",
+            str(Path(sample_ndpi)),
+            "--mode",
+            "save",
+        ],
+    )
```
John-P, line 28: Any particular reason why you are creating a Cli handler and then calling that rather than just using the slide_thumbnail method on WSIReader? Just seems like an odd way to make a thumbnail for the test.

tiatoolbox/visualization/tileserver.py
```diff
@@ -47,7 +51,39 @@ def __init__(self, title: str, layers: Dict[str, WSIReader]) -> None:
...
         self.tia_title = title
-        self.tia_layers = layers
+
+        # generic layer names if none provided
+        if isinstance(layers, list):
+            layers = {f"layer-{i}": p for i, p in enumerate(layers)}
+        # set up the layer dict
+        layer_def = {}
+        meta = None
+        for i, key in enumerate(layers.keys()):
```
John-P, line 61: @measty This could be better written as just `enumerate(layers)` as this will enumerate the keys. This should have been flagged with pre-commit. Are you running pre-commit before committing?

tiatoolbox/visualization/tileserver.py
```diff
@@ -47,7 +51,39 @@ def __init__(self, title: str, layers: Dict[str, WSIReader]) -> None:
...
+            layer = layers[key]
+
+            if isinstance(layer, (str, Path)):
+                p = Path(layer)
```
John-P, line 65: p is not a good variable name. I think this should have been flagged by pre-commit. I suggest something descriptive like `layer_path` instead.

tiatoolbox/visualization/tileserver.py
```diff
@@ -47,7 +51,39 @@ def __init__(self, title: str, layers: Dict[str, WSIReader]) -> None:
...
+                    # assume its a low-res heatmap
+                    layer = Image.open(p)
+                    layer = np.array(layer)
+                else:
+                    layer = WSIReader.open(p)
+
+            if isinstance(layer, np.ndarray):
+                if len(layer.shape) == 2:
+                    # single channel, make into rgb with colormap
+                    c_map = cm.get_cmap("jet")
```
John-P, line 76: It would be nice to be able to change the colormap. I'm personally not a big fan of jet as a hardcoded colormap because it is not perceptually uniform. The luminance does not correspond to the value, hence it is not good for colour blind users. If we must have a hardcoded map or a default I suggest "viridis" (the matplotlib default) or "magma". See here for more information:
- https://matplotlib.org/3.5.0/tutorials/colors/colormaps.html
- https://jakevdp.github.io/blog/2014/10/16/how-bad-is-your-colormap/

tiatoolbox/visualization/tileserver.py
```diff
@@ -47,7 +51,39 @@ def __init__(self, title: str, layers: Dict[str, WSIReader]) -> None:
...
+                    im_rgb = (c_map(layer) * 255).astype(np.uint8)
+                    layer = VirtualWSIReader(im_rgb[:, :, :3], info=meta)
+                else:
+                    layer = VirtualWSIReader(p, info=meta)
```
John-P, line 80: This would section could possibly be refactored into a function e.g. `colourise_image` which takes an ndarray and applies a cmap if is is a single channel array, else returns the image unmodified. Then this could be wrapped in VirtualWSIReader. E.g.

```python
if isinstance(layer, (str, Path)):
    p = Path(layer)
      if p.suffix in [".jpg", ".png"]:  # maybe add .webp and .gif here?
          # assume its a low-res heatmap
            layer = Image.open(p)
            layer = np.array(layer)
            layer = colourised_layer(layer, cmap=...)
            layer = VirtualWSIReader(layer, info=meta)
      else:
           layer = WSIReader.open(p)
```

I think this would make it a bit more readable / maintainable. This also does not account for low res TIFFs which we may need to address later.

John-P (changes requested):
Looks good. Just a few suggestions for changes which might improve the code quality / maintainability in my opinion.

tiatoolbox/cli/tile_server.py
John-P: Is this resolved?

John-P:
Just a side note, can we be consistent with comments capitalisation. I think that we should consistently write comments as a full sentence with a capital letter and full stop unless it is just a word or two note. I recently made a PR to tidy inconsistencies such as this up but I see them creeping in again.

tiatoolbox/visualization/tileserver.py
measty: changed the default to viridis, and can choose the colourmap in colourise function, though there is no provision for choosing colourmap in cli at the moment. Also addressed the other code maintainablility/readability issues above

tests/test_tileserver.py
measty: i think it was a historical artifact of something i tried and removed, cant quite remember now! but have changed it

tiatoolbox/visualization/tileserver.py
measty: moved the colourising code into a function in utils/visualisation. Haven't added any additional formats for now, can maybe look at that later

tiatoolbox/utils/visualization.py
```diff
@@ -33,6 +34,19 @@ def random_colors(num_colors, bright=True):
     return colors
 
 
+def colourise_image(img, cmap="viridis"):
+    """If input img is single channel, colourise it
+    using specified colourmap.
+    """
```
John-P, line 40: Just minor formatting. The first line of the docstring should be a single line with a full stop.

Additional detail is then in a paragraph after a blank line. Also missing args and returns here.

tiatoolbox/cli/tile_server.py
```diff
@@ -0,0 +1,25 @@
...
+from tiatoolbox.cli.common import tiatoolbox_cli, cli_img_input, cli_name
+
+
+@tiatoolbox_cli.command()
+@cli_img_input(multiple=True)
+@cli_name(multiple=True)
+def tile_server(img_input, name):  # pragma: no cover
+    """show a slide, optionally together with one or more whole slide
+    overlays
+    """
```
John-P, line 11: This docstring formatting needs fixing.

tiatoolbox/visualization/tileserver.py
```diff
@@ -2,15 +2,17 @@
...
+from PIL import Image
 
 from tiatoolbox import data
 from tiatoolbox.tools.pyramid import ZoomifyGenerator
-from tiatoolbox.wsicore.wsireader import WSIReader
+from tiatoolbox.wsicore.wsireader import VirtualWSIReader, WSIReader
+from tiatoolbox.utils.visualization import colourise_image
 
 
 class TileServer(Flask):
```
John-P, line 18: Might need to add to the docstring here to explain new behaviour, e.g. how layer names are determined and coloursiation etc.

John-P (changes requested):

John-P:
@measty Can you please mark the comments as resolved if they have been addressed.

PR #322, opened 2022-03-24
MAINT: Add Checks To User Provided Paths

John-P:
This is to satisfy Deepsource security issues and also avoid the user from accidentally overwriting files or writing to a directory that doesn't exists.

### DIFF
--- a/tests/test_annotation_stores.py
+++ b/tests/test_annotation_stores.py
@@ -1,3 +1,4 @@
+  # skipcq: PTC-W6004
 """Tests for annotation store classes."""
 import json
 import pickle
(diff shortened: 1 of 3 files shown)


John-P:
This has triggered a bunch of false positives on DeepSource for the same issue. They are all in test files where the path is controlled by pytest. However, the option to ignore / report it has disappeared for me.

shaneahmed (approved):

shaneahmed (approved):
Thanks @John-P

John-P:
@shaneahmed  Oh this wasn't actually supposed to be merged. There was an issue where deepsource stopped running on it so it appeared to pass checks but that was only because not test were running on it!

PR #328, opened 2022-04-01
ADD: Add Support For Multiple Channel Images (HxWxC)

vqdang:
- Support for resizing multiple channel images
- Support for multiple dtypes conversion


tests/test_utils.py
```diff
@@ -46,14 +46,49 @@ def test_imresize():
     )
     assert resized_img.shape == (1000, 500, 3)
 
-    # test for dtype conversion
+    # test for dtype conversion, pairs of
+    # (original type, converted type)
+    test_dtypes = [
```
shaneahmed, line 51: I do not understand how you define these pairs?

tiatoolbox/utils/transforms.py
```diff
@@ -119,16 +119,48 @@ def imresize(img, scale_factor=None, output_size=None, interpolation="optimise")
         else:
             interpolation = "area"
 
-    if img.dtype == np.float16:
-        img = img.astype(np.float32)
+    # dtypes that cv2.resize can work on out-of-the-box
+    dtype_mapping = [
```
shaneahmed, line 123: why this mapping?

tiatoolbox/utils/transforms.py
```diff
@@ -119,16 +119,48 @@ def imresize(img, scale_factor=None, output_size=None, interpolation="optimise")
...
+        (np.bool, np.uint8),
+        (np.int8, np.int16),
```
shaneahmed, line 125: why int8 to int16?

tests/test_utils.py
vqdang: Opencv resize function only works on a fixed set of np.dtype . Hence, for dtypes that cv2 do not work on, we need to convert it to the closest dtype. Here, the converted dtypes are selected such that it includes the precision range of original dtype. As for how we know which dtype is supported by opencv, I manually run for each dtype and check which one will crash.

tiatoolbox/utils/transforms.py
vqdang: Updated description in https://github.com/TissueImageAnalytics/tiatoolbox/pull/328/commits/5b68415f24b1aeda108f5473f66f8a1f81d0bfab

tiatoolbox/utils/transforms.py
vqdang: Updated description in https://github.com/TissueImageAnalytics/tiatoolbox/pull/328/commits/5b68415f24b1aeda108f5473f66f8a1f81d0bfab

tests/test_utils.py
```diff
@@ -46,14 +46,49 @@ def test_imresize():
...
+        (np.bool, np.uint8),
+        (np.int8, np.int16),
```
shaneahmed, line 53: I understand why you are doing this. Why np.int8 converted to np.int16 ? Is np.int16 not supported in opencv?

vqdang: `np.int8` is not supported but `np.int16` is supported. Hence, `np.int8` (original type) will be converted to `np.int16` (converted type).

shaneahmed (approved):

tests/test_utils.py
```diff
@@ -46,14 +46,49 @@ def test_imresize():
...
+            interpolation=cv2.INTER_CUBIC,
+        )
+        assert resized_img.shape == (50, 50, 3)
+        assert resized_img.dtype == converted_dtype
+
+    # test resizing multiple channels
+    img = np.random.randint(0, 256, (4, 4, 16))
     resized_img = utils.transforms.imresize(
-        img.astype(np.float16),
-        scale_factor=0.5,
```
John-P, line 52: Just spotted that scale factor was changed from 0.5 to 4. Is that correct?

vqdang: It has been refactored into L66 and L67

John-P (approved):
Looks good to me.

PR #330, opened 2022-04-06
NEW: Annotation Rendering

measty:
Draft PR, its still a WIP but it is functional so putting it here for potential feedback. Adds an AnnotationTileGenerator and AnnotationRenderer which allows serving of tiles rendered directly from an annotation store.

For example:

```python
store = SQLiteStore.from_geojson(geo_path)
wsi = WSIReader.open(slide_path)
#renderer will use the value of properties[score_prop] for each annotation to determine color, using
#the matplotlib 'jet' colormap in this case
renderer = AnnotationRenderer(score_prop='probability', mapper='jet')
tile_gen=AnnotationTileGenerator(info=wsi.info, store=store, renderer=renderer)

app = TileServer(
     title='testing server',
     layers={
     "slide": wsi,
     "overlay": tile_gen,
     },
 )
#will set up a tileserver to render annotations as  an overlay onto the wsi
app.run(threaded=False)
```


John-P:
The overall structure looks sensible to me. Nice work! Just wondering how it performs when zoomed out and rendering a lot of polygons. Perhaps we could render them as simple boxes or points when zoomed far out if this becomes a performance bottleneck?

measty:
> The overall structure looks sensible to me. Nice work! Just wondering how it performs when zoomed out and rendering a lot of polygons. Perhaps we could render them as simple boxes or points when zoomed far out if this becomes a performance bottleneck?
> 
> I will do a more thorough review shortly.

It does get a bit slow with a whole slides worth of cells. I have some local changes i'm working on to improve performance when zoomed out, including rendering stuff as boxes when zoomed out. Needs to take size into account too though, as larger annotations will want to still be rendered fully when zoomed out

John-P:
The SQLiteStore keeps track of the bounding box of each annotation directly in the database so it's possible to very efficiently query for just that if you know that's all you need (and could query again only for large annotations). I could add a base function to support this for all subclasses of the base class if that would be helpful here.

measty:
> The SQLiteStore keeps track of the bounding box of each annotation directly in the database so it's possible to very efficiently query for just that if you know that's all you need (and could query again only for large annotations to get the full shape). I could add a base function to support this for all subclasses of the base class if that would be helpful here.

I actually tried playing around with adding that a couple of days ago, i did it by a flag to query which changes the select string:

def query(
        self,
        geometry: QueryGeometry,
        where: Union[str, bytes, Callable[[Geometry, Dict[str, Any]], bool]] = None,
        geometry_predicate: str = "intersects",
        bbox_only: bool = False,
    ) -> List[Annotation]:
        if bbox_only:
            sel_str="properties, cx, cy, min_x, min_y, max_x, max_y"
        else:
            sel_str="properties, cx, cy, geometry"
        query_geometry = geometry
        cur = self._query(
            sel_str,
      etc...
and a few changes to unpack_geometry to handle getting just the bounding box. Do you think it would be better to do it by adding a separate function?

measty:
Another thing I was looking into doing would be to do some sort of decimation when zoomed out, so if a tile has a bunch of small annotations in then some % of them are rendered and the rest ignored - doesn't make much sense to render them all at a zoom level where you wont be able to see individual annotations anyway

ghadjigeorghiou:
> Another thing I was looking into doing would be to do some sort of decimation when zoomed out, so if a tile has a bunch of small annotations in then some % of them are rendered and the rest ignored - doesn't make much sense to render them all at a zoom level where you wont be able to see individual annotations anyway

@measty I agree with you. There is no need to render annotations that won't be visible. It would be nice if you can show the a summary (number of annotations) for that tile when zoomed out. A good example of this is from this screenshot 👇 that a 3rd year student of @shaneahmed created.

Screenshot 2022-04-12 at 10 37 05

When the user is zoomed out you show the summary of the area and while they are zooming in you break that down for different tiles and then you render the individual annotations.  There are 2 example of how to achieve this with Openlayers, but I am not sure if you are using that.

Examples
1. [Clustered Features](https://openlayers.org/en/latest/examples/cluster.html)
2. [Dynamic Clusters](https://openlayers.org/en/latest/examples/clusters-dynamic.html)

John-P:
Just merged in develop. Should fix the travis issue.

John-P:
@measty is this ready to be ready to review (no longer a draft)?

tiatoolbox/annotation/storage.py
```diff
@@ -951,7 +951,7 @@ def pquery(
             >>> store.pquery("props['class']")
             ... {42, 123}
 
-        """
+        """  # noqa
```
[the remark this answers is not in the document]
John-P: Looks like others have this issue also: https://github.com/pgjones/flake8-sql/issues/20

measty:
The problem with the DeepFeatureExtractor test seems to have been caused by my putting the fixtures cell_grid, points_grid and fill_store into conftest.py (I did this so I could reuse them in different test files)

Moving them out of there and making individual copies of them in the test files where they are used seems to have fixed the problem, test is passing now.

I have absolutely no idea why that would have any impact on the DeepFeatureExtractor test, (especially given the test ran fine locally), but for what its worth that seems to have been the cause.

tiatoolbox/annotation/storage.py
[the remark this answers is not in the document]
measty: To be honest I would vote for removing the flake8-sql. It seems to be causing more problems than it solves. Its now causing pre-commit to crash when pushing to git.

Pre-commit hooks run fine locally, and just flags up a few complexity based complaints.

Other than this pre-commit issue with flake-sql, its now ready to be made into a non-draft as everything else is green apart from a couple of complexity flags.

tiatoolbox/annotation/storage.py
```diff
@@ -1628,6 +1628,71 @@ def _append(self, key: str, annotation: Annotation, cur: sqlite3.Cursor) -> None
             token,
         )
 
+    @staticmethod
+    def __initialize_query_string_parameters(
```
John-P, line 1632: This shouldn't be a double underscore

tiatoolbox/annotation/storage.py
```diff
@@ -1628,6 +1628,71 @@ def _append(self, key: str, annotation: Annotation, cur: sqlite3.Cursor) -> None
...
+        query_geometry, query_parameters, geometry_predicate, columns, where
+    ):
+        """Initialises the query string and parameters."""
```
John-P, line 1635: Missing docstring description and args etc.

tiatoolbox/annotation/storage.py
```diff
@@ -1843,6 +1853,40 @@ def bquery(
             }
         return {key: bounds for key, *bounds in cur.fetchall()}
 
+    @staticmethod
+    def __handle_callable_pquery(unique, select, cur, where):
```
John-P, line 1857: Should be single underscore

tiatoolbox/annotation/storage.py
```diff
@@ -1843,6 +1853,40 @@ def bquery(
...
+                if where(json.loads(properties))
+            }
+        return {
+            key: select(json.loads(properties))
+            for key, properties in cur.fetchall()
+            if where(json.loads(properties))
+        }
+
+    @staticmethod
+    def __check_pquery_type(select, where):
```
John-P, line 1872: Should be single underscore

tiatoolbox/annotation/storage.py
```diff
@@ -1843,6 +1853,40 @@ def bquery(
...
+        """Check which type of query it is (str, pickle, callable, star)."""
+        callable_query = any(isinstance(x, Callable) for x in (select, where) if x)
+        pickle_query = any(isinstance(x, bytes) for x in (select, where) if x)
+        str_query = any(isinstance(x, str) for x in (select, where) if x)
+
+        return callable_query, pickle_query, str_query
+
+    @staticmethod
+    def __check_select_where_type(select, where):
```
John-P, line 1881: Should be single underscore

tiatoolbox/annotation/storage.py
```diff
@@ -1843,6 +1853,40 @@ def bquery(
             }
         return {key: bounds for key, *bounds in cur.fetchall()}
 
+    @staticmethod
+    def __handle_callable_pquery(unique, select, cur, where):
+        """Handles Callable pquery."""
```
John-P, line 1858: No explanation of what "handling" is here. No args or returns etc.

tiatoolbox/annotation/storage.py
shaneahmed: Resolved in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/330/commits/82155397526452234dc3fbbfa80a77e520506d13

tiatoolbox/annotation/storage.py
shaneahmed: Resolved in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/330/commits/82155397526452234dc3fbbfa80a77e520506d13

tiatoolbox/annotation/storage.py
shaneahmed: Resolved in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/330/commits/82155397526452234dc3fbbfa80a77e520506d13

tiatoolbox/annotation/storage.py
shaneahmed: @measty and @John-P 
Please can you help with this comment?

tests/test_annotation_stores.py
```diff
@@ -89,11 +89,38 @@ def cell_polygon(
...
+
+    Checks for a class = 123.
+    """
     return props.get("class") == 123
 
 
+def sample_select(props: Dict[str, Any]) -> Tuple[Any]:
+    """Simple example select expression for tests.
+
+    Gets the class value.
```
shaneahmed, line 111:
```suggestion
    Gets the class value.
    
```

tests/test_annotation_stores.py
```diff
@@ -89,11 +89,38 @@ def cell_polygon(
...
+    """
+    return props.get("class")
+
+
+def sample_multi_select(props: Dict[str, Any]) -> Tuple[Any]:
+    """Simple example select expression for tests.
+
+    Gets the class value and the class mod 2.
```
shaneahmed, line 119:
```suggestion
    Gets the class value and the class mod 2.
    
```

tiatoolbox/tools/pyramid.py
```diff
@@ -430,3 +433,234 @@ def tile_path(self, level: int, x: int, y: int) -> Path:
...
+            rgb (np.ndarray):
+                The image to render the annotation on.
+            ann (Annotation):
+                The annotation to render.
+            tl (Tuple[int, int]):
+                The top left coordinate of the tile.
+            scale (int):
+                The scale at which we are rendering the tile.
+            poly_as_box (bool):
+                Whether to render polygons as boxes.
```
shaneahmed, line 612:
```suggestion
                Whether to render polygons as boxes.
                
```

tiatoolbox/utils/visualization.py
```diff
@@ -478,3 +480,194 @@ def to_int_tuple(x):
...
+    score_fn (Callable):
+        an optional callable which will be called on the value of
+        the property that will be used to generate the color before giving
+        it to colormap. Use it for example to normalise property
+        values if they do not fall into the range [0,1], as matplotlib
+        colormap expects values in this range. i.e roughly speaking
+        annotation_color=mapper(score_fn(ann.properties[score_prop]))
+    max_scale (int):
+        downsample level above which Polygon geometries on crowded
+        tiles will be rendered as a bounding box instead
```
shaneahmed, line 514:
```suggestion
        tiles will be rendered as a bounding box instead
        
```

tiatoolbox/utils/visualization.py
```diff
@@ -478,3 +480,194 @@ def to_int_tuple(x):
...
+        return np.squeeze(((np.array(coords) - top_left) / scale).astype(np.int32))
+
+    def get_color(self, ann: Annotation):
+        """Get the color for an annotation.
+        Args:
+            ann (Annotation):
+                Annotation to get color for.
+        Returns:
+            tuple:
+                A color tuple (rgba).
```
shaneahmed, line 565:
```suggestion
                A color tuple (rgba).
                
```

tiatoolbox/utils/visualization.py
```diff
@@ -478,3 +480,194 @@ def to_int_tuple(x):
...
+        """Render a polygon annotation onto a tile using cv2.
+        Args:
+            rgb (ndarray):
+                The tile to render onto.
+            ann (Annotation):
+                The annotation to render.
+            top_left (tuple):
+                The top left corner of the tile in wsi.
+            scale (int):
+                The zoom scale at which we are rendering.
```
shaneahmed, line 590:
```suggestion
                The zoom scale at which we are rendering.
                
```

tiatoolbox/utils/visualization.py
```diff
@@ -478,3 +480,194 @@ def to_int_tuple(x):
...
+        """Render a point annotation onto a tile using cv2.
+        Args:
+            rgb (ndarray):
+                The tile to render onto.
+            ann (Annotation):
+                The annotation to render.
+            top_left (tuple):
+                The top left corner of the tile in wsi.
+            scale (int):
+                The zoom scale at which we are rendering.
```
shaneahmed, line 637:
```suggestion
                The zoom scale at which we are rendering.
                
```

tiatoolbox/utils/visualization.py
```diff
@@ -478,3 +480,194 @@ def to_int_tuple(x):
...
+        """Render a line annotation onto a tile using cv2.
+        Args:
+            rgb (ndarray):
+                The tile to render onto.
+            ann (Annotation):
+                The annotation to render.
+            top_left (tuple):
+                The top left corner of the tile in wsi.
+            scale (int):
+                The zoom scale at which we are rendering.
```
shaneahmed, line 664:
```suggestion
                The zoom scale at which we are rendering.
                
```

tiatoolbox/utils/visualization.py
```diff
@@ -478,3 +478,111 @@ def to_int_tuple(x):
...
+        cv2.drawContours(rgb, [cnt], 0, col, -1)
+
+    def render_rect(self, rgb, ann, tl, scale):
+        """render a box annotation onto a tile using cv2"""
+        col = self.get_color(ann)
+        box = self.to_tile_coords(np.reshape(ann.geometry.bounds, (2, 2)), tl, scale)
+        cv2.rectangle(rgb, box[0, :], box[1, :], col, thickness=-1)
+
+    def render_pt(self, rgb, ann, tl, scale):
+        """render a point annotation onto a tile using cv2"""
```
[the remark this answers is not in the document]
measty: Improved these docstrings as suggested, and a few other ones also

tiatoolbox/annotation/storage.py
```diff
@@ -1592,13 +1767,100 @@ def _query(
...
+            query_geometry = Polygon.from_bounds(*query_geometry)
+
+        if isinstance(where, Callable):
+            columns = callable_columns
+
+        query_parameters = {}
+
+        if isinstance(columns, bytes):
+            query_parameters["select"] = columns
+            columns = "pickle_expression(:select, properties) "
```
measty, line 1844: This is not being covered by tests. As far as I can tell, its not actually reachable because a pickled select is handled elsewhere in the _handle_pickle_callable_pquery, and a pickled select can't ever be handed to _query.

I think we can remove, unless I have missed something

measty: pushed a commit to remove it

tiatoolbox/tools/pyramid.py
```diff
@@ -430,3 +433,251 @@ def tile_path(self, level: int, x: int, y: int) -> Path:
...
+
+        Args:
+            rgb (np.ndarray):
+                The image to render the annotation on.
+            bound_geom (Polygon):
+                A polygon representing the bounding box of the tile.
+            scale (int):
+                The scale at which we are rendering the tile.
+            top_left (Tuple[int, int]):
+                The top left coordinate of the tile.
```
John-P, line 655: I'm a bit unclear as to why both `bound_geom` and `top_left` are given here. Should one not be derived from the other? Is this intended to be able to render an arbitrary region? If so then why is the top_left needed? Is it intended to render a specific tile? If so then why not take in `x, y, level` as with the get tile function?

tiatoolbox/tools/pyramid.py
```diff
@@ -430,3 +433,251 @@ def tile_path(self, level: int, x: int, y: int) -> Path:
...
+            else:
+                r.render_poly(rgb, ann, top_left, scale)
+        elif "Line" in geom_type:
+            r.render_line(rgb, ann, top_left, scale)
+        else:
+            warnings.warn("Unknown geometry")
+
+    def render_annotations(
+        self,
+        rgb: np.ndarray,
```
John-P, line 638: 1. Does this have to be RGB or will it also work with greyscale or RGBA?
2. I could call this image instead of RGB, as RGB sounds like you are supposed to give a colour to be e.g. `(255, 0, 255)`.

tiatoolbox/tools/pyramid.py
```diff
@@ -430,3 +433,251 @@ def tile_path(self, level: int, x: int, y: int) -> Path:
...
+        bounds = locsize2bounds(coord, [self.tile_size * scale] * 2)
+        bound_geom = Polygon.from_bounds(*bounds)
+        rgb = self.render_annotations(rgb, bound_geom, scale, coord)
+
+        return Image.fromarray(rgb)
+
+    def render_by_type(
+        self,
+        rgb: np.ndarray,
+        ann: Annotation,
```
John-P, line 602: Why not just call this `annotation` to be more clear?

tiatoolbox/tools/pyramid.py
measty: It is redundant. I can change that.

tiatoolbox/tools/pyramid.py
measty: It will work with rgb or rgba. It wouldnt work with grayscale.

I called it rgb for consistency with pyramid/zoomify generator.
Can change it to image if desired but would be best to change it in both places in that case.

tiatoolbox/tools/pyramid.py
measty: sure, it can be changed to annotation

tiatoolbox/tools/pyramid.py
measty: Have removed the redundant top_left, renamed ann to annotation and renamed rgb to tile (the function is get_tile so I thought it made most sense to have the thing that is being passed around and eventually returned be called tile too)

shaneahmed (approved):

Issue #325, opened 2022-03-30
Bug in tiatoolbox/examples/inference-pipelines/idars.ipynb


DavidBAEpstein:
* TIA Toolbox version: 1.0.1
* Python version: 3.7.12
* Operating System: Macos 11.6.4

### Conclusions before Evidence (see Lewis Carroll _Sentence first—verdict afterwards_)
1. `PatchPredictor` needs to be rewritten so that its member function `predict` cannot contradict whether GPU is in use or not.
2. The example notebooks have no unittests. But these notebooks are supposed to be the most used part of TIA Toolbox. It's not very nice if a person's first contact with our package is a string of crashes. However, I know nothing about tests and nothing about how to use GPUs. How difficult would it be to test the running of each notebook under different environments, in particular with and without GPU?
3. In any case, it would be wise for all the example notebooks to be tested manually without GPU, just to make sure. I (David) can do that.

### Description
*Describe what you were trying to get done.*
I was testing idars.ipynb without GPU
*Tell us what happened, what went wrong, and what you expected to happen.*
Fatal errors. I expected the jupyter notebook to run to conclusion, but slowly

### What I Did
First I corrected one fatal error, caused by ON_GPU being undefined. (The user had been instructed in a text cell to skip the cell defining ON_GPU if not using a GPU.) Then I ran it again, and there was a different fatal error. _**The notebook seems to have been tested only with GPU.**_ The code cell causing the error had 3 statements, so I split the cell into three, and realised that the code was not set up to work without GPU.

The lines
```
tumour_predictor = PatchPredictor(
    pretrained_model='resnet18-idars-tumour',
    batch_size=64,
    num_loader_workers=8)
```
expect a GPU, but the subsequent lines
```
tumour_output = tumour_predictor.predict(
    imgs=wsi_file_list,
    mode='wsi',
    return_probabilities=True,
    on_gpu=ON_GPU)
```
do not expect a GPU, because at this point, `ON_GPU == False`

*Paste the command(s) you ran and the output.* 
I kept pressing Command Return, which on a Mac runs the next cell.

*If there was a crash, please include the traceback here.*
```
RuntimeError                              Traceback (most recent call last)
~/opt/miniconda3/envs/tiatoolbox/lib/python3.7/site-packages/torch/utils/data/dataloader.py in _try_get_data(self, timeout)
   1010         try:
-> 1011             data = self._data_queue.get(timeout=timeout)
   1012             return (True, data)

[...]

PR #331, opened 2022-04-08
DOC: IDaRS Algorithm

DavidBAEpstein:
This branch was started in order to include, in a text cell, pseudo-code for the IDaRS algorithm. The algorithm actually used in the experiments described in the paper by Bilal et al is included here. One reason this is important is that the algorithm listed in the Supplementary Materials for Bilal et al is not correct, and does not correctly describe what was actually done in that work.

Running the notebook on a Macbook revealed a bug, resulting in a crash, as described in Issue #325. This is now fixed, and the notebook completes overnight on a Macbook. Clean-up of files downloaded by the notebook has also been improved.

Many minor changes have been made to text cells, to make them clearer or eliminate unnecessary words.


DavidBAEpstein:
I finally got 'make html' in docs to work, and found that readthedocs version of the text cells was horrible, although it looks fine with the standard notebook rendering. I converted to draft in order to use sphinx in the jupyter notebook.

PR #358, opened 2022-06-10
BUGFIX: Fixing for Merging Patch Predictor Output

### LINKED ISSUE #356: Bug: Incorrect output when return_raw=True in predictor.merge_predictions() and overlapping happens
* TIA Toolbox version: 1.1.0
* Python version: 3.7.13
* Operating System: Linux

### Description

Hello everyone,

I attempted to create a mask of the probability outputs for a whole slide image after prediction with the resnet18-kather100k pretrained model. The WSI predictions were made with stride = 100.

After loading the json file of the saved predictions, I passed the appropriate wsi_output through predictor.merge_predictions() with **return_raw=True**. The returned prediction map probabilities appear to be incorrectly scaled. For any single prediction, the probabilities of all classes should add up to 1 and that was not the case for the returned output maps. After Investigating I believe I found the bug in the code below. When looping through the coordinates of the patch predictions, each box is being populated with the predictions. Since stride(100) < patch_size(224) there is some overlap in this case. Currently, the overlapped elements are being overwritten by the current prediction and then being divided by the denominator array. I think the overlapping predictions should be added to instead of overwriting so that dividing by the denominator determines the average. 

Referenced Section:
https://github.com/TissueImageAnalytics/tiatoolbox/blob/1e808d3c846a64fc3ac9fcec7da460bd3ca21e15/tiatoolbox/models/engine/patch_predictor.py#L344-L357

Line with Bug:
https://github.com/TissueImageAnalytics/tiatoolbox/blob/1e808d3c846a64fc3ac9fcec7da460bd3ca21e15/tiatoolbox/models/engine/patch_predictor.py#L351

Proposed Fix: 
`output[tl[1] : br[1], tl[0] : br[0]] += prediction`

### What I Did

```
pred_map = predictor.merge_predictions(wsi_file_name, wsi_output, 
                                           resolution=overview_resolution, 
                                           units=overview_unit, return_raw=True)

np.sum(pred_map[1000,1000,:]) # Sum of probabilities of all classes for a single point in prediction map
Out[3]: 0.16666667194728055
```

vqdang:
To resolve #356

### DIFF
--- a/tiatoolbox/models/engine/patch_predictor.py
+++ b/tiatoolbox/models/engine/patch_predictor.py
@@ -348,7 +348,7 @@ class PatchPredictor:
             tl = np.ceil(np.array(bound[:2]) * fx).astype(np.int32)
             # bot-right for output placement
             br = np.ceil(np.array(bound[2:]) * fx).astype(np.int32)
-            output[tl[1] : br[1], tl[0] : br[0]] = prediction
+            output[tl[1] : br[1], tl[0] : br[0]] += prediction
             if denominator is not None:
                 denominator[tl[1] : br[1], tl[0] : br[0]] += 1
 
--- a/tiatoolbox/models/engine/patch_predictor.py
+++ b/tiatoolbox/models/engine/patch_predictor.py
@@ -42,7 +42,7 @@ class IOPatchPredictorConfig(IOSegmentorConfig):
 
 
 class PatchPredictor:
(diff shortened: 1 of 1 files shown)


shaneahmed (commented):
Need to fix deepsource error.

shaneahmed (approved):
Thanks @vqdang

PR #369, opened 2022-06-17
DEV: Fix Flake8 Errors in `utils/transforms.py`

shaneahmed:
- Fix CCR001 Cognitive complexity is too high
- Fix R504 unecessary variable assignement before return statement


tiatoolbox/utils/transforms.py
```diff
@@ -83,7 +83,7 @@ def imresize(img, scale_factor=None, output_size=None, interpolation="optimise")
     """
     if scale_factor is None and output_size is None:
         raise TypeError("One of scale_factor and output_size must be not None.")
-    if scale_factor is not None:
+    elif scale_factor:
```
John-P, line 86: Why elif? this would never be reached if raise is hit. I think this should be if not elif.

tiatoolbox/utils/transforms.py
```diff
@@ -94,7 +94,7 @@ def imresize(img, scale_factor=None, output_size=None, interpolation="optimise")
         height = int(img.shape[0] * scale_factor[1])
         output_size = (width, height)
 
-    if scale_factor is None:
+    elif scale_factor is None:
```
John-P, line 97: This could be just else

shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/369/commits/aec15a336f84a63afce520b51b2faa1f43b00853

PR #374, opened 2022-06-20
DEV: Add New Pre-Commit Hooks

John-P:
The only file that has been edited is `.pre-commit-config.yaml`.
Adds new pre-commit hooks:

- check-ast               # Simply checks whether the files parse as valid Python.
- fix-byte-order-marker   # Removes utf-8 byte order marker.
- check-json              # Checks JSON files for parseable syntax.
- check-merge-conflict    # Checks for files that contain merge conflict strings.
- check-symlinks          # Checks for symlinks which do not point to anything.
- check-toml              # Checks TOML files for parseable syntax.
- check-xml               # Checks XML files for parseable syntax.
- check-yaml              # Checks YAML files for parseable syntax.
- destroyed-symlinks      # Detects symlinks which are changed to regular files with a content of a path which that symlink was pointing to.
- detect-private-key      # Detects the presence of private keys.
- end-of-file-fixer       # Ensures that a file is either empty, or ends with one newline.
- mixed-line-ending       # Replaces or checks mixed line endings.
- requirements-txt-fixer  # Sorts entries in requirements.txt.
- trailing-whitespace     # Trims trailing whitespace.
  - exclude: 'setup.cfg'
- file-contents-sorter    # Sorts the lines in specified files (defaults to alphabetical).
  - files: whitelist.txt
- black-jupyter    # Formats Jupyter notebook code with black.
- debug-statements        # Checks for debugger imports and py37+ `breakpoint()` calls in python source. 
- pretty-format-json      # Sets a standard for formatting JSON files. 
- rst-backticks               # Detect common mistake of using single backticks when writing rst.
- rst-directive-colons    # Detect mistake of rst directive not ending with double colon.
- rst-inline-touching-normal # Detect mistake of inline code touching normal text in rst.


shaneahmed (approved):
Thanks @John-P. 
I will merge this after I have fixed a few other issues related to existing pre-commit errors and random failures on travis.

vqdang (commented):
As long as nothing broken then it would be ok. Im a bit concerned about the notebook. Could you check if the doc generated correctly with images for both the pdf and the web ?

PR #380, opened 2022-06-22
DEV: Replace PlattScaler With Sklearn LogisticRegression

John-P:
Replace the custom PlattScaler in `tools/scale.py` with the regular Scikit-Learn LogisticRegression.

Notebook needs full test run to confirm the change is ok for the notebook.
I cannot currently download the images because TCGA is down. Does anyone have some data (clinical CSV and SVS files) to run this? @vqdang ?


shaneahmed (approved):
Thanks @John-P If it runs fine on the notebook, then I am fine with this.

PR #384, opened 2022-06-29
BUG: Fix Errors After Torch Upgrade To Version 1.12

shaneahmed:
- Fix Errors after torch upgrade to v1.12
The issue seems to be due change in output dtype in torch v1.12 which affects precision https://github.com/pytorch/pytorch/releases


John-P (approved):

vqdang:
With the drastic changes in CUDA and GPU operations, it may be necessary to prevent installing the latest pytorch version or allow installing w/o pytorch (i.e users will handle that) as there will be more breaking changes going on.

As the performance of predictive test are still within the error margin, this quick reduction of precision criteria for **features** is kind of ok.

vqdang (approved):

PR #386, opened 2022-07-01
DEV: Set Up GitHub Actions Workflow (Build, Test & Release)

John-P:
This PR adds support for running continuous integration on GitHub Actions, including:
- Running Tests
- Reporting coverage (DeepSource & Codecov)
- Releasing to PyPI (and Test PyPI)

It currently uses the beta Ubuntu 22.04 image and runs against python versions: 3.8, 3.9, and 3.10.


tests/models/test_multi_task_segmentor.py
```diff
@@ -12,7 +12,7 @@
 ON_GPU = toolbox_env.has_gpu()
 # The value is based on 2 TitanXP each with 12GB
 BATCH_SIZE = 1 if not ON_GPU else 16
-NUM_POSTPROC_WORKERS = 2 if not toolbox_env.running_on_travis() else 8
+NUM_POSTPROC_WORKERS = 2 if not toolbox_env.running_on_ci() else 8
```
shaneahmed, line 15: Can we try to increase `NUM_POSTPROC_WORKERS` as this is the limit on travis. May be GitHub actions provides more.

tests/models/test_semantic_segmentation.py
```diff
@@ -33,7 +33,7 @@
 ON_GPU = toolbox_env.has_gpu()
 # The value is based on 2 TitanXP each with 12GB
 BATCH_SIZE = 1 if not ON_GPU else 16
-NUM_POSTPROC_WORKERS = 2 if not toolbox_env.running_on_travis() else 8
+NUM_POSTPROC_WORKERS = 2 if not toolbox_env.running_on_ci() else 8
```
shaneahmed, line 36: Same as above.

shaneahmed (commented):
Thanks @John-P Looks great. Just one test if we can try to increase NUM_POSTPROC_WORKERS.

tests/models/test_multi_task_segmentor.py
John-P: Looks like a two core CPU virtual machine with 7GB RAM: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

John-P: Also, why not just use the `multiprocessing.cpu_count()` here instead of a hard-coded value of 2 or 8?

John-P: Or better yet: `len(os.sched_getaffinity(0))`

shaneahmed: Fine with me

John-P:
@shaneahmed I believe this is ready. The coverage is showing as -0.01%, but I can't see why other than a potential rounding issue (we could possibly set a threshold in the codecov config for this, e.g. allow a decrease of <0.5% to avoid this).

tests/models/test_multi_task_segmentor.py
John-P: I will address this on another PR.

shaneahmed (approved):
Thanks @John-P 
This will help a lot in development of the toolbox.

PR #344, opened 2022-05-09
DEV: Depreciate Support For Python 3.7.

shaneahmed:
- Remove support for python 3.7 to allow using latest version of image codecs and some advanced features.

PR #391, opened 2022-07-04
REV: Revert changes from #344

shaneahmed:
- Revert changes from #344

### DIFF
--- a/tests/test_annotation_stores.py
+++ b/tests/test_annotation_stores.py
@@ -1033,9 +1033,9 @@ class TestStore:
             store._load_cases(["foo"], lambda: None, lambda: None)
 
     @staticmethod
-    def test_py38_init(fill_store, store_cls, monkeypatch):
+    def test_py37_init(fill_store, store_cls, monkeypatch):
         """Test that __init__ is compatible with Python 3.8."""
-        py38_version = (3, 8, 0)
+        py37_version = (3, 7, 0)
 
         class Connection(sqlite3.Connection):
             """Mock SQLite connection."""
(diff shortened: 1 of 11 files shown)


tests/test_annotation_stores.py
```diff
@@ -1033,9 +1033,9 @@ def test_load_cases_error(fill_store, store_cls):
             store._load_cases(["foo"], lambda: None, lambda: None)
 
     @staticmethod
-    def test_py38_init(fill_store, store_cls, monkeypatch):
+    def test_py37_init(fill_store, store_cls, monkeypatch):
         """Test that __init__ is compatible with Python 3.8."""
```
John-P, line 1037: Minor comment inconsistency

John-P, line 1037:
```suggestion
        """Test that __init__ is compatible with Python 3.7."""
```

tests/test_annotation_stores.py
```diff
@@ -1033,9 +1033,9 @@ def test_load_cases_error(fill_store, store_cls):
...
-        py38_version = (3, 8, 0)
+        py37_version = (3, 7, 0)
```
John-P, line 1038: Interestingly, I think this should never have been 38 but should be fixed to 3.7/ Was it changed when deprecating 3.7 before? I think the aim of this test is to test for lack of the deterministic key word, which is missing before 3.8 (<= 3.7).

John-P (approved):
Generally looks good. Couple of minor points. I am kind of surprised that this works given that imagecodecs dropped 3.7, but I guess that our constraints must contain a version before that happened.

EDIT: Ah yes, we specify that the last version to support 3.7 is ok (https://pypi.org/project/imagecodecs/2021.11.20/).

tests/test_annotation_stores.py
shaneahmed: Yes. I just manually reverted #344

tests/test_annotation_stores.py
shaneahmed: Resolved in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/391/commits/ecf47c6532b90ad72524838e9e65f0dd11f35f06

PR #392, opened 2022-07-04
TST: Add Test To Check Requirements Consistency

John-P:


John-P:
Another test which might be worth adding as a GitHub action is checking that the conda environment resolves in a reasonable amount of time, e.g. 10 minutes.

tests/test_requirements.py
```diff
@@ -0,0 +1,192 @@
+from pathlib import Path
+from typing import Dict, List, Tuple
+
+import pytest
+import yaml
+from pkg_resources import Requirement
+
+REQUIREMENTS_FILES = [
```
DavidBAEpstein, line 8: Should requirements.win64.conda.yml be part of the picture of inclusions and constraints?

John-P: Thanks for catching this, I've added this in [f195d96](https://github.com/TissueImageAnalytics/tiatoolbox/pull/392/commits/f195d96e64a4542ca5543898f3637bd947c1855b)

DavidBAEpstein (approved):
Good idea to do this static test, which should be very quick. Should the Windows requirements file be compared with either of the other main files?

John-P:
> 



> Good idea to do this static test, which should be very quick. Should the Windows requirements file be compared with either of the other main files?

I currently just compare it in the global comparison (for packages which are in all files). However, it could potentially be compared with the regular dev file. We currently don't have a windows dev file so I have left it as None for now.

DavidBAEpstein:
Iff conda is flailing around, the probability that dev and main end up inconsistent will grow. (I mean a python program may succeed with main environment but fail with dev environment, even though the constraints are the same.)

An off-topic, but related, point is: do non-developers really need all those sphinx packages clogging up their small machines? If the answer is “Yes” then I would respond that, for simplicity and ease of maintenance, dev and main should be identical.) There doesn’t seem to be any point in having two different requirements files, if no attempt is made to keep one of them small.


Best wishes,
David



On 4 Jul 2022, at 18:09, John Pocock ***@***.******@***.***>> wrote:



Another check which might be worth adding as a GitHub action is checking that the conda environment resolves in a reasonable amount of time, e.g. 10 minutes.

DavidBAEpstein:
I didn’t read the code carefully. Why the emphasis on “all”. I would have thought that, any time there is a key in common between two or more of the five requirements files, the constraints should be the same. But I’m out of my depth and don’t really know what I’m talking about.

David

On 4 Jul 2022, at 18:58, John Pocock ***@***.******@***.***>> wrote:



Good idea to do this static test, which should be very quick. Should the Windows requirements file be compared with either of the other main files?

I currently just compare it in the global comparison (for packages which are in all files). However, it could potentially be compared with the regular dev file. We currently don't have a windows dev file so I have left it as None for now.

tests/test_requirements.py
```diff
@@ -0,0 +1,202 @@
...
+    ("requirements.win64.conda.yml", None),
+]
+
+
+def parse_pip(
+    file_path: Path = None, lines: List[str] = None
+) -> Dict[str, Requirement]:
+    """Parse a pip requirements file.
+
+    Not that package names are case insensitive and underscores and
```
shaneahmed, line 21:
```suggestion
    Note that package names are case insensitive and underscores and
```

John-P:
@DavidBAEpstein I have adjusted the test to check for consistency when a key appears in more than one file, as per your suggestion.

John-P:
@shaneahmed The windows requirements file has some oddities (a mix of dev and non dev dependencies). Additionally, there are several things in the conda files e.g. h5py which we do not directly use and could probably be removed.

shaneahmed:
> @shaneahmed The windows requirements file has some oddities (a mix of dev and non dev dependencies). Additionally, there are several things in the conda files e.g. h5py which we do not directly use and could probably be removed.

Let's make it consistent.

shaneahmed:
I think we can update `albumentations>=1.2.0`

shaneahmed:
Shall we include `docs/requirements.txt` in dev checks?

John-P:
> Let's make it consistent.

Consistent with regular conda requirements?

shaneahmed:
How about `setup.py`. In v1.2.0, flask is missing.

John-P:
> How about `setup.py`. In v1.2.0, flask is missing.

If we move to using requirements.txt in setup.py then it will be checked by this test.

John-P:
@shaneahmed I've added a special case for setup.py

John-P:
@shaneahmed and @DavidBAEpstein, This has now been converted to a pre-commit hook instead of a unit test.

shaneahmed:
I guess it just checks version numbers only not missing dependencies.
I added flask to setup.py to make a commit and noticed this.

John-P:
> I guess it just checks version numbers only not missing dependencies.
> 
> I added flask to setup.py to make a commit and noticed this.

Yes, it would not be possible to statically check for missing dependencies without first querying pip for all of the packages to find the dependants of those etc.

PR #398, opened 2022-07-05
DEV: Add GitHub Action To Perform Conda Solve

John-P:
This PR adds a GitHub action which will attempt to solve the conda env for Ubuntu and Windows. It will run:
- When a requirements*.yml file is pushed.
- On the 1st of the month at midnight (on develop).

It will time out if the environment is not solved in 10 minutes.

Windows conda currently takes longer than 10 minutes and so the timeout may need to be increased, or there may be some constraints which need adjusting there.


John-P:
This took a long time to get working but it appears to be ok now.

shaneahmed:
> This took a long time to get working, but it appears to be ok now. @shaneahmed do you know why the windows solve is taking a lot longer (>10m vs 6m for Ubuntu)?

openslide may be?

shaneahmed (approved):
Thanks @John-P 

Looks Good!

PR #403, opened 2022-07-06
BUG: Fix tileserver CLI Issues

measty:
- [x] Clarify help string for cli options where multiple instances can be provided.
- [x] Add flask as requirement to setup.py
- [x] Fix missing index.html while launching through CLI.


shaneahmed:
Can we make Args consistent with the Options
MicrosoftTeams-image (12)

John-P:
I think 'tile-server' is not a clear name for the command for a user to be honest. Surely something like 'show-wsi' would be more clear.

John-P:
I've also noticed that it allows you to launch without specifying a path to an image. This should throw an error.

John-P:
> * Fix missing index.html while launching through CLI.

I am unable to reproduce this. It returns 200 and loads just fine for me. Can you give me details on how to reproduce this?

measty:
> * I've also noticed that it allows you to launch without specifying a path to an image. This should throw an error.
> * How can a user give a list of WSIs? I would have thought they should be able to give a list to have them show as layers.
> * `img-input` is not a nice option name, `imange`, `input` or even `input-image` would be better, It reads very weirdly being 'image input'. Really they should be positional IMO, but if you are going to use an option it should be easy to type and have both a short and long form e.g. `-n` and `--name` etc.

Back when I made this cli I had defined my own custom option name, and was told it should instead use the ones defined in cli\common.py if possible. Hence it was changed to use img-input. 
A user can give multiple layers by supplying multiple instances of the --img-input option, for example:

tiatoolbox tile-server --img-input path\to\slide1 --name slide --img-input path\to\overlay1 --name overlay

measty:
> > * Fix missing index.html while launching through CLI.
> 
> I am unable to reproduce this. It returns 200 and loads just fine for me. Can you give me details on how to reproduce this?

This I think only appears when tiatoolbox is installed through pip, the folder which contains the index.html doesn't end up being copied over even though it exists in the tiatoolbox git

John-P:
> index.html doesn't end up being copied over even though it exists in the tiatoolbox git.

Ah! That makes sense. It needs to be in the manifest to be copied over.

tiatoolbox/cli/common.py
```diff
@@ -32,14 +32,18 @@ def cli_img_input(
...
     return click.option("--img-input", help=usage_help, type=str, multiple=multiple)
 
 
 def cli_name(
-    usage_help: str = "user defined name to be used as an identifier",
+    usage_help: str = "User defined name to be used as an identifier.",
     multiple: bool = False,
 ) -> callable:
     """enables --name option for cli"""
+    if multiple:
```
shaneahmed, line 45: This line is not fully covered in coverage. Please add a test for False condition to cover this line.

John-P:
> A user can give multiple layers by supplying multiple instances of the --img-input option

I'm surprised that click doesn't add something to the description to indicate this. We should state this, else a user would not know.

measty:
> > A user can give multiple layers by supplying multiple instances of the --img-input option
> 
> I'm surprised that click doesn't add something to the description to indicate this. We should state this, else a user would not know.
Yes, ive addressed that in the changes  to help. It'll now say in the help:


--img-input TEXT Path to WSI or directory containing WSIs. Multiple
instances may be provided.
--name TEXT User defined name to be used as an identifier. Multiple
instances may be provided.

If multiple are allowed

tiatoolbox/cli/common.py
measty: I do not know how to test a click.option in isolation at the moment. If I figure it out I will add it.

measty: Now done

tests/test_tileserver.py
```diff
@@ -91,3 +92,22 @@ def test_create_with_dict(sample_svs):
         response = client.get("/layer/Test/zoomify/TileGroup0/0-0-0.jpg")
         assert response.status_code == 200
         assert response.content_type == "image/jpeg"
+
+
+def test_cli_name_multiple_flag():
+    """Test cli_name multiple flag."""
+    opt = cli_name()
```
John-P, line 99: Any particular reason this is being renamed? It could be confusing to see `@opt` if this was missed higher up.

measty: No, it was just a relic of something else I tried. Have removed the renaming.

PR #430, opened 2022-07-25
BUG: Correct error in nucleus instance segmentor engine

adamshephard:
Correct error in nucleus instance segmentor engine where for tiles of different size widths/heights there are boundary artefacts.


vqdang (approved):
Dumbest mistake ever made by me, thanks Mark.

PR #431, opened 2022-07-26
BUG: Update HoVer-Net+ Post-processing

adamshephard:
- Fix an issue with the HoVer-Net+ post-processing. Previously, a simple blur was applied for the post-processing. Now it is updated inline with the original work - using various morphological operations specific to each epithelial layer. 

- Remove warnings from importing specific libraries in both HoVer-Net and HoVer-Net+.


John-P:
Looks good. Just a small note on performance though. I think the cv2 equivalent operations might be a bit faster than the skimage ones. This is minor but could have a bigger impact on a WSI level.

shaneahmed:
Looks good to me. I agree with John's comment on using opencv for morphological processing.

adamshephard:
> Looks good. Just a small note on performance, though. I think the cv2 equivalent operations might be a bit faster than the skimage ones. This is minor but could have a bigger impact on a WSI level.
> 
> See https://opencv24-python-tutorials.readthedocs.io/en/latest/py_tutorials/py_imgproc/py_morphological_ops/py_morphological_ops.html#morphological-ops

Hi John, after looking at this I see that there are no opencv equivalents for the functions that I have used. I've also looked at the source code for these functions, and they rely on scipy anyway. Probably best to keep as it is?

shaneahmed (approved):

PR #436, opened 2022-07-29
:construction: WIP: Add Ability To Apply Affine Transform When Reading Zarr

John-P:
This is a work in progress.

- [x] Investigate application of transforms when reading numpy/scipy array-like objects.
- [x] Test with a zarr array.
- [x] Profile speed impact (about 10x slower than without when using order 1, AKA linear, interpolation).
- [ ] Integrate with tiatoolbox API e.g. `NGFFWSIReader` or separate wrapper object.
- [ ] Investigate application of generic coordinate transforms. I.E. can this be structured in a way for minimal changes to do both?
- [ ] Document bug/variance discovered between numpy and zarr indexing. E.g. `array[..., :3]` works for numpy but not zarr, zarr must use `array[:, :, :3]`. This broke things and took a while to debug/discover.
- [ ] Remove image dumping from tests.

# Demo

## Order 1 (Linear)

This is with order 1 (linear) interpolation, which appears to be the best balance of appearance and performance.

test_affine_zarr_view

## Order 0 (Nearest Neighbour)
test_affine_zarr_view_order_0

## Order 2 (Quadratic)
test_affine_zarr_view_order_2

## Order 3 (Cubic)
test_affine_zarr_view_order_3


John-P:
Closed due to inactivity. Can re-open later if it becomes a priority.

PR #439, opened 2022-08-05
DOC: Improve 07-advanced-modeling.ipynb

DavidBAEpstein:
- DOC: Edit metadata to make all cells visible to rtd.
- MAINT: edit raw source of list for jnb-rtd consistency.
- MAINT: Remove redundant initial ">" and incorrect double quotes. Then edit for clarity.
- MAINT: change "temp" to "tmp"
- MAINT: change "tail --lines 1" to "tail -n 1". Change wording around ON_GPU.
- MAINT: correct typo
- BUG: Remove pretence that ON_GPU=False is viable/meaningful with current software packages.
- MAINT: Run Colab with GPU and make requisite changes.
- MAINT: Change number of WORKERS to 4. Add comment that the notebook crashes if run without Cuda-enabled GPU.


examples/07-advanced-modeling.ipynb
```diff
@@ -94,11 +94,11 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
-    "ON_GPU = True  # Set `ON_GPU=True` if cuda-enabled GPU is availabe,\n",
+    "ON_GPU = True  # Set `ON_GPU` ti `True` if cuda-enabled GPU is available,\n",
```
shaneahmed, line 101: `ON_GPU` to `True`?

DavidBAEpstein: Since the notebook crashes without GPU, I rewrote this saying that ON_GPU = False is not allowed. Some other notebooks have been written so that they still run successfully, though slowly, without GPU. But that is not the case here.

DavidBAEpstein:
I have checked the rendering of this jupyter notebook against the rendering provided by readthedocs. These two renderings are as similar as it is reasonable to require. The PR has been in a reasonable state for approval for some time, and I think it should now be approved by the reviewers.

shaneahmed (approved):
Thanks @DavidBAEpstein

PR #445, opened 2022-08-19
ENH: Annotation Storage Additions For Visualization

measty:
This PR adds some enhancements to storage.py that are useful when using an AnnotationStore for visualization purposes.

Main changes:

Support for adding annotations from a hovernet-style .dat through a utility function

Addition of 'area' field to table in SQliteStore. Annotations are returned sorted by area, and supports querying for annotations satisfying an area threshold, useful for zoomed-out rendering where annotations which are too small to display can be excluded.
After the discussions in tiatoolbox meeting, functions are added to allow to add/remove the area column to save space if its not expected to be needed.
Sorting annotations efficiently by area (can create an index on it) allows to effficiently deal with potentially overlapping annotations by drawing geometries of smallest area last. Eg

Unsorted rendering:
before_area_sort

Sorted rendering:
after_area_sort

~~Addition of cacheable versions of query, bquery. The intention with these is to potentially be more efficient when viewing the same area of screen and modifying various options controlling how annotations are rendered (perhaps changing color mapping, or deselecting objects of specific types to be renderered). As in these cases we are displaying the same annotations just in a different way, we dont need to re-load all the annotations for each tile from the store; we can just load the cached annotations and then filter/render them differently.
Still needs a bit of testing to see how much this actually helps/if its worth using~~ (caching aspects of the PR removed for now after feedback, may revisit this later in a separate PR)


tiatoolbox/annotation/storage.py
```diff
@@ -1321,19 +1557,22 @@ def clear(self) -> None:
...
+                e.g. ":memory:" or "./data.db"), a pathlib Path, or a file
+                handle.
+            path (Path):
+                The path to the annotation store data. This will be
+                ":memory:" if the annotation store is in-memory. This is
+                derived from `connection` and normalised to be a pathlib
+                Path object.
+            con (sqlite3.Connection):
+                The sqlite3 database connection.
+    <<<<<<< HEAD
```
shaneahmed, line 1572: Delete 

<<<<<<< HEAD
    =======
    >>>>>>> b4cf2ebc5ad0c367d425c50b631748de42434e4f

tiatoolbox/annotation/storage.py
```diff
@@ -1303,6 +1522,23 @@ def to_dataframe(self) -> pd.DataFrame:
         )
         return pd.json_normalize(features).set_index("key")
 
+    def translate_db(self, x: float, y: float):
+        """Translate all annotations by the given vector. Useful for
```
shaneahmed, line 1526:
```suggestion
        """Translate all annotations by the given vector. 
        
        Useful for
```

tiatoolbox/annotation/storage.py
```diff
@@ -1895,20 +2154,226 @@ def query(
...
+        rows: str,
+        geometry: Optional[Iterable] = None,
+        callable_rows: Optional[str] = None,
+        geometry_predicate="intersects",
+        con=None,
+        bbox=True,
+        compress_type=None,
+        min_area=None,
+    ) -> sqlite3.Cursor:
+        """Common query construction logic for `cached_query` and `cached_bquery`. Similar
```
shaneahmed, line 2187:
```suggestion
        """Common query construction logic for `cached_query` and `cached_bquery`. 
        
        Similar
```

tiatoolbox/annotation/storage.py
```diff
@@ -1895,20 +2154,226 @@ def query(
...
+            for properties, cx, cy, blob in cur.fetchall()
+        ]
+
+    def cached_bquery(
+        self,
+        geometry: Optional[Tuple] = None,
+        where: Callable[[Dict[str, Any]], bool] = None,
+        min_area=None,
+    ) -> Dict[str, Annotation]:
+        """Query the store for annotations. Similar to `bquery` but
```
shaneahmed, line 2306:
```suggestion
        """Query the store for annotations. 
        
        Similar to `bquery` but
```

tiatoolbox/annotation/storage.py
```diff
@@ -1895,20 +2154,226 @@ def query(
...
+            return data
+        return {key: ann for key, ann in data.items() if where(ann.properties)}
+
+    def cached_query(
+        self,
+        geometry: Optional[Tuple] = None,
+        where: Callable[[Dict[str, Any]], bool] = None,
+        min_area=None,
+    ) -> List[Annotation]:
+        """Query the store for annotations. Similar to `query` but
```
shaneahmed, line 2345:
```suggestion
        """Query the store for annotations. 
        
        Similar to `query` but
```

tiatoolbox/annotation/storage.py
```diff
@@ -1136,19 +1142,232 @@ def _load_cases(
...
+                replaced by the corresponding value. Useful for providing descriptive
+                names to non-descriptive types,
+                eg {1: 'Epithelial Cell', 2: 'Lymphocyte', 3: ...}.
+            relative_to [float, float]:
+                The x and y coordinates to use as the origin for the annotations.
+
+        Returns:
+            AnnotationStore:
+                A new annotation store with the annotations loaded from the file.
+        """
```
shaneahmed, line 1205:
```suggestion

        """
```

tiatoolbox/annotation/storage.py
```diff
@@ -1136,19 +1142,232 @@ def _load_cases(
...
```
shaneahmed, line 1172:
```suggestion

        """
```

John-P:
I will have to provide longer feedback in future but just making a few quick notes before our meeting. I have a few concerns:
1. I don't think it wise to add more columns to the database. When doing the extensive analysis and use case argument for using this SQLite store, storage space was a big concern, especially with large sets of annotations. I would suggest using a temporary table to keep areas for visualization purposes. This will compute them once when the database is opened and not write them to disk, taking up more space. Alternatively, you could use the bounding box area which can be obtained in a select statement from the rtree index already.
2. Beware caching. It can be full of pitfalls from memory leaks to cache invalidation issues. The built in functools LRU cache has been known to have issues with memory leaks. I would suggest moving this up the stack out of the storage class itself and into the visualization layer also. There is the issue of if someone update the database from another process. How does this get propagated etc.

measty:
> 



> I will have to provide longer feedback in future, but just making a few quick notes before our meeting. I have a few concerns:
> 
> 1. I don't think it wise to add more columns to the database. When doing the extensive analysis and use case argument for using this SQLite store, storage space was a big concern, especially with large sets of annotations. I would suggest using a temporary table to keep areas (and other region props) for visualization purposes. This will compute them once when the database is opened and not write them to disk, taking up more space. Alternatively, you could use the bounding box area, which can be obtained in a select statement from the rtree index already without the need to extra columns. Additionally, I think that area might be calculated already when the WKB is parsed. Have you checked this? I.e. does it offer any speedup to have this saved somehwere vs obtaining it from the shapely object?
> 2. Beware caching. It can be full of pitfalls from memory leaks to cache invalidation issues. The built-in functools LRU cache has been known to have issues with memory leaks. I would suggest moving this up the stack out of the storage class itself and into the visualization layer also. There is the issue of if someone updates the database from another process. How does this get propagated etc.
> 3. Dat is a very generic format. Calling it `from_dat` is confusing as it is specific to our toolbox / hovernet rather than being able to use any dat file. Perhaps `from_hovernet_dat` would be better. Also, this adds the first tight coupling between the annotation store and another part of the toolbox. I think we should avoid this. Especially, if annotation store is to be broken out into a separate package in the future. I can't think of an elegant solution right now but perhaps some loader function which takes a store and a dat file to load into the database would be better than baking it into the storage class to avoid this coupling for now.

1. I have changed the area column to an interger to save space, as that is sufficient for what is needed. The impact on filesize was around 3-4%. My concerns with the alternate solutions are that BBox area is not sufficient, as that gives a large area to line annotations, which really need their correct zero area as they will show up well plotted over a region annotation, but not vice-versa. And the temporary table solution, I am concerned that calculating the area of all annotations and creating an index on it on load will lengthen the already noticable delay on opening a new store. My first attempt at using area originally had all the filtering/sorting by area outside of the storage class, I moved it into the db as its more efficient to have it there so that you can have an index on it, and so area constraints can be incorporated directly in the query so that you dont even return from the database the annotations you dont plan to render.
2. I've removed the cached querys for now, can potentially rethink how its done later. 
3. I've moved the code to create a store from hovernet style .dat into a standalone function in tiatoolbox\utils\misc.py

measty:
I've added functions to add/drop the area column as discussed. Also finished adding tests, so marked it as ready for review.

John-P:
@measty It has since occurred to me that we may need to add a column for annotation type (e.g. an integer which corresponds to an enum) for filtering annotation etc. This could also be used instead of / as a fallback for area. I.e.g points/lines get 0 (no computation needed) and polygons get bounding box from the rtree at query. This should be very fast as it is a simple multiplication performed by the query engine. It may in fact be more performant than using the actual area on cold startup and have negligible difference for warm queries also. Can we benchmark this?

tests/test_annotation_stores.py
```diff
@@ -755,6 +838,36 @@ def test_from_geojson_path(fill_store, tmp_path, store_cls):
...
+        com2 = annotations_center_of_mass(list(store2.values()))
+        assert com2.x == pytest.approx((com.x - 100) * 2)
+        assert com2.y == pytest.approx((com.y - 100) * 2)
+
+    @staticmethod
+    def test_transform_db(fill_store, tmp_path, store_cls):
+        """Test translating a store."""
+
+        def test_translation(geom):
+            """Performas a translation of input geometry."""
```
shaneahmed, line 861:
```suggestion
            """Performs a translation of input geometry."""
```

tests/test_utils.py
```diff
@@ -1366,3 +1369,104 @@ def test_detect_gpu():
...
+def test_from_dat(tmp_path):
+    """Test generating an annotation store from a .dat file."""
+    data = make_simple_dat()
+    joblib.dump(data, tmp_path / "test.dat")
+    store = utils.misc.store_from_dat(tmp_path / "test.dat")
+    assert len(store) == 2
+
+
+def test_from_dat_typedict(tmp_path):
+    """Test generating an annotation store from a .dat file with a typedict."""
```
shaneahmed, line 1396:
```suggestion
    """Test generating an annotation store from a .dat file with a type dict."""
```

tests/test_utils.py
```diff
@@ -1366,3 +1369,104 @@ def test_detect_gpu():
...
+        tmp_path / "test.dat", scale_factor=2, relative_to=(50, 50)
+    )
+    result = store.query(where="props['type'] == 1")
+    # check centroid is at 150,150
+    poly = next(iter(result.values()))
+    assert np.rint(poly.geometry.centroid.x) == 150
+    assert np.rint(poly.geometry.centroid.y) == 150
+
+
+def test_from_multihead_dat(tmp_path):
```
shaneahmed, line 1420:
```suggestion
def test_from_multihead_dat(tmp_path):
```

tests/test_utils.py
```diff
@@ -1366,3 +1369,104 @@ def test_detect_gpu():
...
+    """Test generating an annotation store from a .dat file with multiple heads."""
+    head_a = make_simple_dat()
+    head_b = make_simple_dat([(200, 200), (300, 300)])
+    data = {
+        "A": head_a,
+        "B": head_b,
+        "resolution": 0.5,
+        "othermetadata": {"foo": "bar"},
```
shaneahmed, line 1428:
```suggestion
        "other_meta_data": {"foo": "bar"},
```

tiatoolbox/annotation/storage.py
```diff
@@ -1117,18 +1118,84 @@ def _load_cases(
         raise IOError("Invalid file handle or path.")
 
     @classmethod
-    def from_geojson(cls, fp: Union[IO, str]) -> "AnnotationStore":
-        geojson = cls._load_cases(
+    def from_geojson(
+        cls,
+        fp: Union[IO, str],
+        scale_factor=1,
```
shaneahmed, line 1124:
```suggestion
        scale_factor: int=1,
```

tiatoolbox/annotation/storage.py
```diff
@@ -1117,18 +1118,84 @@ def _load_cases(
...
+        relative_to=None,
```
shaneahmed, line 1125:
```suggestion
        relative_to:Tuple[float, float]=None,
```

tiatoolbox/annotation/storage.py
```diff
@@ -1117,18 +1118,84 @@ def _load_cases(
...
+        """
+        store = cls()
+        store.add_from_geojson(fp, scale_factor, relative_to=relative_to)
+        return store
+
+    def add_from_geojson(
+        self,
+        fp: Union[IO, str],
+        scale_factor=1,
+        relative_to=None,
```
shaneahmed, line 1151: Add typehints

tiatoolbox/annotation/storage.py
```diff
@@ -1117,18 +1118,84 @@ def _load_cases(
...
             file_fn=json.load,
         )
-        store = cls()
-        for feature in geojson["features"]:
-            geometry = feature2geometry(feature["geometry"])
-            properties = feature["properties"]
-            store.append(Annotation(geometry, properties))
-        return store
+
+        anns = [
```
shaneahmed, line 1187:
```suggestion
annotations = [
```

tiatoolbox/annotation/storage.py
```diff
@@ -1318,7 +1401,6 @@ class SQLiteMetadata(MutableMapping):
             Path object.
         con (sqlite3.Connection):
             The sqlite3 database connection.
```
shaneahmed, line 1403:
```suggestion
new line
"""
```

tiatoolbox/utils/misc.py
```diff
@@ -849,3 +854,157 @@ def select_cv2_interpolation(scale_factor):
     if np.any(scale_factor > 1.0):
         return "cubic"
     return "area"
+
+
+def store_from_dat(
+    fp: Union[IO, str],
```
shaneahmed, line 860: Add type hints

tiatoolbox/utils/misc.py
```diff
@@ -849,3 +854,157 @@ def select_cv2_interpolation(scale_factor):
...
+    scale_factor=1,
+    typedict=None,
+    relative_to=None,
+    cls: AnnotationStore = SQLiteStore,
+) -> "AnnotationStore":
+    """Load annotations from a hovernet-style .dat file.
+    Args:
```
shaneahmed, line 867: new line

tiatoolbox/utils/misc.py
```diff
@@ -849,3 +854,157 @@ def select_cv2_interpolation(scale_factor):
...
+        return poly
+    warnings.warn("Invalid geometry found, fix using buffer().")
+    return poly.buffer(0.01)
+
+
+def anns_from_hoverdict(data, props, typedict, relative_to, scale_factor):
+    """Helper function to create list of Annotation objects from a
+    hovernet-style dict of segmentations, mapping types using typedict
+    if provided.
+    """
```
shaneahmed, line 913: new line

tiatoolbox/utils/misc.py
```diff
@@ -849,3 +854,157 @@ def select_cv2_interpolation(scale_factor):
...
+        # transform coords to be relative to given pt.
+        poly = translate(poly, -relative_to[0], -relative_to[1])
+    if poly.is_valid:
+        return poly
+    warnings.warn("Invalid geometry found, fix using buffer().")
+    return poly.buffer(0.01)
+
+
+def anns_from_hoverdict(data, props, typedict, relative_to, scale_factor):
+    """Helper function to create list of Annotation objects from a
```
shaneahmed, line 910: first line should be sentence
add help for args.

tiatoolbox/utils/misc.py
```diff
@@ -849,3 +854,157 @@ def select_cv2_interpolation(scale_factor):
...
+            },
+        )
+        for ann in data.values()
+    ]
+
+
+def make_default_dict(data, subcat):
+    """Helper function to create a default typedict if none is provided.
+
+    The unique types in the data are given a prefix to differentiate
```
shaneahmed, line 940: Args:?

tiatoolbox/utils/misc.py
```diff
@@ -849,3 +854,157 @@ def select_cv2_interpolation(scale_factor):
...
+            with a type that is a key in the dictionary, will have their type
+            replaced by the corresponding value. Useful for providing descriptive
+            names to non-descriptive types,
+            eg {1: 'Epithelial Cell', 2: 'Lymphocyte', 3: ...}.
+            For multi-head output, should be a dict of dicts, eg:
+            {'head1': {1: 'Epithelial Cell', 2: 'Lymphocyte', 3: ...},
+                'head2': {1: 'Gland', 2: 'Lumen', 3: ...}, ...}.
+        relative_to [float, float]:
+            The x and y coordinates to use as the origin for the annotations.
+    """
```
shaneahmed, line 982: new line
returns:?

measty: That function doesnt return anything

shaneahmed: ok. please can you add new line?

measty:
> Thanks @measty. Please can you address the comments?
@shaneahmed the docstring-related comments have been addressed

measty:
> @measty It has since occurred to me that we may need to add a column for annotation type (e.g. an integer which corresponds to an enum) for filtering annotations etc. This could also be used instead of / as a fallback for area. E.g points/lines get 0 (no computation needed) and polygons get bounding box from the rtree at query. This should be very fast as it is a simple multiplication performed by the query engine. It may in fact be more performant than using the actual area on cold startup (because calculating actual area is slow vs fast bounding box) and may have negligible difference for warm queries also. Can we benchmark this?
> 
> If you still want the actual area, the bounding box could be used if the area column is missing at start up (or first query) to reduce lag time. Meanwhile, in the background you could perform a query to add the missing column to avoid start up lag. However, we may find that this combo of annotation type + bounding box is enough.

@John-P I tried adding a custom function:

```
def get_area_approx(objtype, min_x, min_y, max_x, max_y):
       """Function to get the approx. area of a geometry."""
       if objtype in ("Point", "LineString"):
           return 0
       return (max_x - min_x) * (max_y - min_y)
```

And using that for the area based sorting/filtering. I did a small bit of benchmarking, rendering all the possible tiles in the generator, and just the first 50 to simulate its effect after initial load. 
All tiles: 279s using above function, 275s using area column
First 50: 4.7s using above fn, 3.06 using area column.

So, it seems to be a reasonable fallback, and doesn't degrade performance much apart from a bit of a hit to the initial lag.

tests/test_annotation_stores.py
```diff
@@ -755,6 +849,36 @@ def test_from_geojson_path(fill_store, tmp_path, store_cls):
         store2 = store_cls.from_geojson(tmp_path / "polygon.json")
         assert len(store) == len(store2)
 
+    @staticmethod
+    def test_from_geojson_path_transform(fill_store, tmp_path, store_cls):
+        """Test loading from geojson with a transform"""
```
shaneahmed, line 854:
```suggestion
        """Test loading from geojson with a transform."""
```

tests/test_annotation_stores.py
```diff
@@ -755,6 +849,36 @@ def test_from_geojson_path(fill_store, tmp_path, store_cls):
...
+        )
+        assert len(store) == len(store2)
+        com2 = annotations_center_of_mass(list(store2.values()))
+        assert com2.x == pytest.approx((com.x - 100) * 2)
+        assert com2.y == pytest.approx((com.y - 100) * 2)
+
+    @staticmethod
+    def test_transform_db(fill_store, tmp_path, store_cls):
+        """Test translating a store."""
+
```
shaneahmed, line 870:
```suggestion
```

tiatoolbox/annotation/storage.py
shaneahmed: please add typehints

tiatoolbox/annotation/storage.py
```diff
@@ -1117,18 +1118,84 @@ def _load_cases(
...
-        geojson = cls._load_cases(
+    def from_geojson(
+        cls,
+        fp: Union[IO, str],
+        scale_factor=1,
+        relative_to=None,
+    ) -> "AnnotationStore":
+        """Create a new database with annotations loaded from a geoJSON file.
+        Args:
+            fp (Union[IO, str, Path]):
```
shaneahmed, line 1129: Please can you check if we are using file_path variable in misc? It would be better to be consistent. Also, fp is not clear.

tiatoolbox/annotation/storage.py
```diff
@@ -1117,18 +1118,84 @@ def _load_cases(
...
+                A new annotation store with the annotations loaded from the file.
+
+        """
+        store = cls()
+        store.add_from_geojson(fp, scale_factor, relative_to=relative_to)
+        return store
+
+    def add_from_geojson(
+        self,
+        fp: Union[IO, str],
```
shaneahmed, line 1149:
```suggestion
        file_path: Union[IO, str],
```

tiatoolbox/annotation/storage.py
shaneahmed: Please be consistent with typehints.

tiatoolbox/annotation/storage.py
```diff
@@ -1117,18 +1118,84 @@ def _load_cases(
...
+        Args:
+            fp (Union[IO, str, Path]):
+                The file path or handle to load from.
+            scale_factor (float):
+                The scale factor to use when loading the annotations. All coordinates
+                will be multiplied by this factor to allow import of annotations saved
+                at non-baseline resolution.
+            relative_to [float, float]:
+                The x and y coordinates to use as the origin for the annotations.
+        """
```
shaneahmed, line 1165:
```suggestion

        """
```

tiatoolbox/annotation/storage.py
```diff
@@ -1117,18 +1118,84 @@ def _load_cases(
...
+
```
shaneahmed, line 1166:
```suggestion
```

tiatoolbox/annotation/storage.py
```diff
@@ -1288,6 +1355,22 @@ def to_dataframe(self) -> pd.DataFrame:
...
+        """Transform all annotations in the store using provided function.
+
+        Useful for transforming coordinates from slide space into
+        patch/tile/core space, or to a different resolution, for example.
+
+        Args:
+            transform (callable[Geometry, Geometry]):
+                A function that takes a geometry and returns a new
+                transformed geometry.
+        """
```
shaneahmed, line 1368:
```suggestion

        """
```

tiatoolbox/annotation/storage.py
```diff
@@ -1318,7 +1401,6 @@ class SQLiteMetadata(MutableMapping):
             Path object.
         con (sqlite3.Connection):
             The sqlite3 database connection.
-
     """
```
shaneahmed, line 1404:
```suggestion

    """
```

tiatoolbox/annotation/storage.py
```diff
@@ -1382,6 +1464,7 @@ def __init__(
         connection: Union[Path, str, IO] = ":memory:",
         compression="zlib",
```
shaneahmed, line 1465: Please be consistent with typehints.

tiatoolbox/annotation/storage.py
```diff
@@ -1500,11 +1600,14 @@ def register_custom_function(
                 cx INTEGER NOT NULL,     -- X of centroid/representative point
                 cy INTEGER NOT NULL,     -- Y of centroid/representative point
                 geometry BLOB,           -- Detailed geometry
-                properties TEXT          -- JSON properties
+                properties TEXT,         -- JSON properties
+                area INTEGER NOT NULL    -- Area (for ordering)
             )
             """
```
shaneahmed, line 1606:
```suggestion

            """
```

tiatoolbox/annotation/storage.py
```diff
@@ -1677,7 +1785,7 @@ def _append(self, key: str, annotation: Annotation, cur: sqlite3.Cursor) -> None
...
                 INSERT INTO annotations VALUES(
                     NULL, :key, :geom_type,
-                    :cx, :cy, :geometry, :properties
+                    :cx, :cy, :geometry, :properties, :area
                 )
                 """,
```
shaneahmed, line 1790:
```suggestion

                """,
```

tiatoolbox/utils/misc.py
```diff
@@ -849,3 +854,196 @@ def select_cv2_interpolation(scale_factor):
...
+
+
+def add_from_dat(
+    store,
+    fp: Union[IO, str],
+    scale_factor=1,
+    typedict=None,
+    relative_to=None,
+) -> None:
+    """Add annotations from a .dat file to an existing store. Make
```
shaneahmed, line 999:
```suggestion
    """Add annotations from a .dat file to an existing store. 
    
    Make a best effort to create valid shapely geometries from provided contours.
```

tiatoolbox/utils/misc.py
```diff
@@ -849,3 +854,196 @@ def select_cv2_interpolation(scale_factor):
...
+    a best effort to create valid shapely geometries from provided contours.
```
shaneahmed, line 1000:
```suggestion
```

shaneahmed (approved):
Looks good to me @measty.

Please make sure that your address the comments. My main comments are on type hints and consistency of variable names. I will wait for @John-P's review before merge.

tiatoolbox/annotation/storage.py
measty: I have used fp in everything I have added, to be consistent with what was already used throughout storage.py

measty: Have replaced fp with full_path

John-P: It's supposed to be a common shorthand for file path here (rather than full path). Happy with file_path instead.

John-P:
One thing to note is that if any changes are being made to the table schema, this must be done carefully to avoid breaking things too much. E.g. we need:
- To document changes in the docstrings.
- To increment the version number.
- To add checks for version number.
- Handle old versions / upgrade if required.

John-P:
Just copying over comments here from out last team chat, If we change the type column to be an int/enum instead of text then this would be a breaking change. However, it could be a significant space-saving and hopefully more than account for adding a true area column if this is required. By using the type, we can also get a pseudo area, i.e. 0 for lines and points and the bounding-box area for polygons. Additionally, we could probably remove the cx and cy columns. They are currently used to store a 'representative point', i.e. a point near the center of mass but within the shape. We are currently not using this. Additionally, a bounding box center point can be obtained from the rtree index.

tests/test_annotation_stores.py
```diff
@@ -15,7 +15,7 @@
 import pandas as pd
 import pytest
 from shapely import affinity
-from shapely.geometry import Polygon
+from shapely.geometry import MultiPoint, Polygon
```
John-P, line 18: Can I just check why you are using MultiPoint here? I personally can't see many reasons to use this over multiple Point annotations as it makes spatial queries not work as well. Is this just to test that they do work?

tiatoolbox/annotation/storage.py
```diff
@@ -1071,80 +1072,147 @@ def to_geodict(self) -> Dict[str, Any]:
...
+    ) -> "AnnotationStore":
+        """Create a new database with annotations loaded from a geoJSON file.
+        Args:
+            full_path (Union[IO, str, Path]):
+                The file path or handle to load from.
+            scale_factor (float):
+                The scale factor to use when loading the annotations. All coordinates
+                will be multiplied by this factor to allow import of annotations saved
+                at non-baseline resolution.
+            relative_to [float, float]:
```
John-P, line 1135:
```suggestion
            origin [float, float] = (0, 0):
```

I think 'origin' is slightly less ambiguous / more clear than 'relative_to' IMO.

tiatoolbox/annotation/storage.py
```diff
@@ -1071,80 +1072,147 @@ def to_geodict(self) -> Dict[str, Any]:
...
         raise IOError("Invalid file handle or path.")
 
     @classmethod
-    def from_geojson(cls, fp: Union[IO, str]) -> "AnnotationStore":
-        geojson = cls._load_cases(
-            fp=fp,
+    def from_geojson(
+        cls,
+        full_path: Union[IO, str],
+        scale_factor: float = 1,
```
John-P, line 1124: Maybe overkill but perhaps list of transforms could be given to allow other things just as inverting the origin / mirroring or rotating etc. Similar to the coordinateTransformations in NGFF v0.4.

tiatoolbox/annotation/storage.py
```diff
@@ -1071,80 +1072,147 @@ def to_geodict(self) -> Dict[str, Any]:
...
+            scale_factor (float):
+                The scale factor to use when loading the annotations. All coordinates
+                will be multiplied by this factor to allow import of annotations saved
+                at non-baseline resolution.
+            relative_to [float, float]:
+                The x and y coordinates to use as the origin for the annotations.
+
+        """
+
+        def transform_geom(geom):
```
John-P, line 1168:
```suggestion
        def transform_geomety(geom):
```

'geom' is not used anywhere else. I think just using the actual word is more clear.

tiatoolbox/annotation/storage.py
```diff
@@ -1071,80 +1072,147 @@ def to_geodict(self) -> Dict[str, Any]:
...
+            """Helper function to transform a geometry if needed."""
+            if relative_to is not None:
+                # transform coords to be relative to given pt.
```
John-P, line 1171:
```suggestion
                # Transform coords to be relative to given point.
```

tiatoolbox/annotation/storage.py
```diff
@@ -1071,80 +1072,147 @@ def to_geodict(self) -> Dict[str, Any]:
...
+    def to_geojson(self, full_path: Optional[IO] = None) -> Optional[str]:
         """Serialise the store to geoJSON.
 
         For more information on the geoJSON format see:
         - https://geojson.org/
         - https://tools.ietf.org/html/rfc7946
 
         Args:
-             fp (IO):
+             full_path (IO):
```
John-P, line 1209: This name doesn't make sense. It is not for a path but a file-like IO object. 
`fp` was chosen to be consistent with the Python standard library e.g. json: https://github.com/python/cpython/blob/c7662420d62e04eefd99d7cdbcc74f72e68f3a6a/Lib/json/__init__.py#L120

Although admittedly this also works with a file path. and should be a union of `IO` and `pathlib.Path`.

Full path also doesn't make sense as it works with non-full paths i.e. relative paths.

John-P, line 1209:
```suggestion
             full_path (IO|pathlib.Path|str):
```

tiatoolbox/annotation/storage.py
```diff
@@ -1170,12 +1238,12 @@ def write_geojson_to_file_handle(file_handle: IO):
             file_handle.write("]}")
 
         return self._dump_cases(
-            fp=fp,
+            full_path=full_path,
```
John-P, line 1241: Again full path doesn't make sense as a name here. It can be an IO object or a file path (full or relative etc.).

tiatoolbox/annotation/storage.py
```diff
@@ -1288,6 +1356,23 @@ def to_dataframe(self) -> pd.DataFrame:
         )
         return pd.json_normalize(features).set_index("key")
 
+    def transform_db(self, transform: Callable[[Geometry], Geometry]) -> None:
```
John-P, line 1359: If this is transforming the store don't put db in the name. Is this being implemented for all stores to keep the API consistent?

John-P, line 1359:
```suggestion
    def transform(self, transform: Callable[[Geometry], Geometry]) -> None:
```

John-P (changes requested):
I've make a few comments / suggestions of things. Mostly just suggesting alternative naming for things.

tiatoolbox/annotation/storage.py
```diff
@@ -1071,80 +1072,147 @@ def to_geodict(self) -> Dict[str, Any]:
...
+                ),
+                feature["properties"],
+            )
+            for feature in geojson["features"]
+        ]
+
+        print(f"added {len(annotations)} annotations")
+        self.append_many(annotations)
+
+    def to_geojson(self, full_path: Optional[IO] = None) -> Optional[str]:
```
John-P, line 1201:
```suggestion
    def to_geojson(self, full_path: Optional[IO|pathlib.Path|str] = None) -> Optional[str]:
```

`|` can be used instead of `Union[IO, str, Path]`, not sure if this is in python 3.7 though so might have to be `Union[IO, str, Path]`

tiatoolbox/annotation/storage.py
```diff
@@ -1371,17 +1456,24 @@ class SQLiteStore(AnnotationStore):
 
     Uses and rtree index for fast spatial queries.
 
+    Version History:
+    1.0.0:
+        Initial version.
+    1.0.1 (07/10/2022):
+        Added support for Area column and queries sorted/filtered by area.
```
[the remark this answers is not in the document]
John-P: Just wanted to clarify how this is handling a store when loaded which does not have an area column> Is it just ignoring it and raising an error if you try to order on area? Or is it adding the area column at load time? If that latter, I am concerned that this would slow down loading of datasets if not using them for visualization.

tiatoolbox/annotation/storage.py
measty: It is implemented in the AnnotionStore base class so will be available in all stores

John-P: Great, thanks for clarifying. I missed that it was in the base class.

tiatoolbox/annotation/storage.py
measty: Origin would be clearer, i'll change it to that

tiatoolbox/annotation/storage.py
measty: its just a change of origin, but worded badly. I'll clarify

tiatoolbox/annotation/storage.py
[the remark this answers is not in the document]
measty: At the moment, the behaviour is that if you try to make a query that filters by area, and there isn't an area column, it raises an error stating that there is no area column.

Ordering by area happens by default if the area column is available. If it's not no ordering happens. At the moment there is no ability to specify what ordering happens. Probably in 2.0 we could add support for specifying what to sort on, be it area or perhaps some property.

tiatoolbox/annotation/storage.py
measty: yes, older versions of python weren't happy with | so will have to be union for now

tiatoolbox/annotation/storage.py
measty: i've changed this back to fp again

measty:
> I've make a few comments / suggestions of things. Mostly just suggesting alternative naming for things.

All the comments have been addressed now, I think

Issue #446, opened 2022-08-22
PatchExtractor filtering


rogertrullo:
* TIA Toolbox version: 1.2.1
* Python version: 3.8
* Operating System:  ubuntu 20

### Description
Hi, thanks for sharing this library!
I was checking the  [PatchExtractor class](https://github.com/TissueImageAnalytics/tiatoolbox/blob/develop/tiatoolbox/tools/patchextraction.py) and it seems that patches containing at least one pixel of positive mask, will be included.
Depending on the scenario this can be undesired e.g. some patches will be mainly background.
I was wondering if we can include some kind of parameter e.g. min number of positive pixels or minimum area of the patch being positive?

shaneahmed:
@mostafajahanifar Please can you look into this?

rogertrullo:
@mostafajahanifar if it can help, I added a simple PR [#451 ](https://github.com/TissueImageAnalytics/tiatoolbox/pull/451) of what it could look like

PR #451, opened 2022-08-28
ENH: Add Min Mask Ratio Parameter in PatchExtractor

rogertrullo:
This PR adds the functionality described in[ #446 ](#446 ), basically adds a `min_mask_ratio` parameter to the patchextractor class so that one can define it as a percentage of the patch area that needs to include mask.


tiatoolbox/tools/patchextraction.py
```diff
@@ -109,6 +116,13 @@ def __init__(
         self.locations_df = None
         self.coordinate_list = None
         self.stride = None
+
+        assert (
```
shaneahmed, line 120: Please raise ValueError. Assert is not allowed in the code.

shaneahmed:
@rogertrullo Please add/modify tests to address coverage.
Please also fix pre-commit and deepsource errors.

tiatoolbox/tools/patchextraction.py
```diff
@@ -280,7 +302,18 @@ def filter_coordinates_fast(
         flag_list = []
         for coord in scaled_coords:
             this_part = tissue_mask[coord[1] : coord[3], coord[0] : coord[2]]
-            flag_list.append(np.any(this_part > 0))
+            patch_area = np.prod(this_part.shape)
+            pos_area = this_part.sum()
+            if min_mask_ratio == 0:
```
mostafajahanifar, line 307: I'm not sure why you considered an special case `min_mask_ratio == 0`.
I believe the whole situation can be rewritten with one `if` statement:
```
if pos_area > patch_area * min_mask_ratio: # note that ">" is used instead of ">="
    flag_list.append(True)
else:
    flag_list.append(False)
```

hopefully, this would also help with coverage and you won't need to add more tests to cover this bit.

tiatoolbox/tools/patchextraction.py
mostafajahanifar: On this, I believe you don't need to check for the type and range here as you are doing this in the `filter_coordinates_fast` function as well where I believe is the right place to do it. So, get rid of this check here and replace the one in `filter_coordinates_fast` with `ValueError`.

Furthermore, you need to add a test for this special case in the unit test, when an unacceptable value for `min_mask_ratio` is passed to the function. The test should be added somewhere around here:  
https://github.com/rogertrullo/tiatoolbox/blob/0ee90054fc4c4b88251be57b868f51d32554ab69/tests/test_patch_extraction.py#L456

mostafajahanifar (changes requested):
Thanks @rogertrullo for doing this important enhancement 👍 

The PR reasoning and solution look pretty straightforward. However, I asked for some changes to improve the readability and functionality of the code. Also, you need to update the tests for the `filter_coordinates_fast` function to improve the coverage your PR.

rogertrullo:
Thanks @mostafajahanifar and @shaneahmed , I was wondering, is the wiki for contributors up to date? I am having a hard time setting up the environment, e.g. when creating a conda env with python 3.7 and pip installing requirements_dev.txt , I am having some package conflicts errors. When using python 3.8 it is fine, however, precommit is failing with `An unexpected error has occurred: CalledProcessError: command: ('conda/envs/tia/bin/python', '-mvirtualenv', 'pre-commit/repooth77l7w/py_env-python3.7', '-p', 'python3.7')`
could you please share more about your workflow and setup?

shaneahmed:
> Thanks @mostafajahanifar and @shaneahmed , I was wondering, is the wiki for contributors up to date? I am having a hard time setting up the environment, e.g. when creating a conda env with python 3.7 and pip installing requirements_dev.txt , I am having some package conflicts errors. When using python 3.8 it is fine, however, precommit is failing with `An unexpected error has occurred: CalledProcessError: command: ('conda/envs/tia/bin/python', '-mvirtualenv', 'pre-commit/repooth77l7w/py_env-python3.7', '-p', 'python3.7')` could you please share more about your workflow and setup?

Hi @rogertrullo , You are right there seems to be some conflict on python 3.7 although we tested this on Friday and it worked fine. A quick fix might be to setup python 3.9 environment and let pre-commit run on pre-commit.ci. You can use `--no-verify` flag for local commits.

tiatoolbox/tools/patchextraction.py
rogertrullo: @mostafajahanifar The issue I found was, when setting min_mask_ratio=1. With that value I would expect to include all patches completely inside the mask. However, if we only use `>` no patch will be included because the maximum value is 1.

shaneahmed:
> > Thanks @mostafajahanifar and @shaneahmed , I was wondering, is the wiki for contributors up to date? I am having a hard time setting up the environment, e.g. when creating a conda env with python 3.7 and pip installing requirements_dev.txt , I am having some package conflicts errors. When using python 3.8 it is fine, however, precommit is failing with `An unexpected error has occurred: CalledProcessError: command: ('conda/envs/tia/bin/python', '-mvirtualenv', 'pre-commit/repooth77l7w/py_env-python3.7', '-p', 'python3.7')` could you please share more about your workflow and setup?
> 
> Hi @rogertrullo , You are right there seems to be some conflict on python 3.7 although we tested this on Friday and it worked fine. A quick fix might be to setup python 3.9 environment and let pre-commit run on pre-commit.ci. You can use `--no-verify` flag for local commits.

@rogertrullo I have checked requirements_dev.txt resolves after applying fix in this PR #455

tiatoolbox/tools/patchextraction.py
mostafajahanifar: Good point about all white mask patches 👍 
Your suggestion would do the job, however, something like below can be more readable which takes into account the special case of `min_mask_ratio=1`:
```
if (pos_area  == patch_area) or (pos_area > patch_area * min_mask_ratio):
    flag_list.append(True)
else:
    flag_list.append(False)
```

rogertrullo: I was checking this test:
https://github.com/rogertrullo/tiatoolbox/blob/0ee90054fc4c4b88251be57b868f51d32554ab69/tests/test_patch_extraction.py#L414-L425
```
    bbox_list = np.array(
        [
            [0, 0, 4, 4],
            [0, 4, 4, 8],
            [4, 0, 8, 4],
            [4, 4, 8, 8],
            [8, 0, 12, 4],
            [8, 4, 12, 8],
        ]
    )
    mask = np.zeros([9, 6])
    mask[0:4, 3:8] = 1  # will flag first 2
```
which after clipping gives this:
```
array([[0, 0, 4, 4],
       [0, 4, 4, 8],
       [4, 0, 6, 4],
       [4, 4, 6, 8],
       [6, 0, 6, 4],
       [6, 4, 6, 8]])
```
which would give the last two patches to be of area=0, and if we add min_area_ratio=0, they will be added incorrectly. Should we add a special condition on area==0?

shaneahmed (approved):
Thanks @rogertrullo 
Looks good to me. I will merge this once @mostafajahanifar has approved it.

mostafajahanifar (approved):
All look good to me as well.
Thanks @rogertrullo  for the great suggestion and good work on this PR 💯 
Looking forward to seeing more contributions from you :-)

PR #456, opened 2022-09-01
DEV: Create pip Install From Git Workflow

John-P:
This adds a GitHub workflow which simply installs openslide and openjpeg binaries and then runs pip install from the github repo which will install the latest develop version. It then checks that `import tiatoolbox` runs without error.


shaneahmed (approved):
Looks good. Please let me know if you want to add this for Windows as well. Otherwise I will merge this.

PR #458, opened 2022-09-03
DOC: Spell Output Correctly

DavidBAEpstein:
Corrects a single typo in a python module


DavidBAEpstein (commented):
A one-character typo

shaneahmed (approved):
Thanks @DavidBAEpstein 
There is a new pre-commit error. I will fix it in a separate PR.

shaneahmed:
> Thanks @DavidBAEpstein There is a new pre-commit error. I will fix it in a separate PR.

@John-P I think PatchExtractor is not written as an abstract class. So we can remove the inheritance from it. Re dsl, this requires `@abstractmethod` decorators.

PR #464, opened 2022-09-09
DEV: Add macOS To pip Install Workflow

John-P:
This PR adds macOS pip install test cases to the GitHub workflow.


shaneahmed:
Please can you add the workflow for checking requirements file as well?
As this will fail once I use requirements.txt in setup.py

John-P:
> Please can you add the workflow for checking requirements file as well?
As this will fail once I use requirements.txt in setup.py

Sorry, I'm not sure what this means. I don't know what requirements check you need here. I cannot see a reason why this would fail when using requirements.txt in setup.py.

shaneahmed:
> > Please can you add the workflow for checking requirements file as well?
> > As this will fail once I use requirements.txt in setup.py
> 
> Sorry, I'm not sure what this means. I don't know what requirements check you need here. I cannot see a reason why this would fail when using requirements.txt in setup.py.

Please check this workflow https://github.com/TissueImageAnalytics/tiatoolbox/blob/develop/.github/workflows/conda-env-create.yml. This requires copying the requirements file to temp directory. This doesn't work on MacOS.

John-P:
> Please check this workflow https://github.com/TissueImageAnalytics/tiatoolbox/blob/develop/.github/workflows/conda-env-create.yml. This requires copying the requirements file to temp directory. This doesn't work on MacOS.

The linked workflow is using a different install process (anaconda). This workflow is just using pip and does not need a tmp directory.

John-P:
@shaneahmed Looks like setup.py doesn't like the commands beginning with "--". Could potentially try simply filtering out those lines.

shaneahmed (approved):
Thanks @John-P

PR #467, opened 2022-09-12
DEV: Use `requirements.txt` in `setup.py`

shaneahmed:
- Use `requirements.txt` in `setup.py`


requirements.txt
```diff
@@ -1,3 +1,5 @@
+# torch installation
+--extra-index-url https://download.pytorch.org/whl/cu113; sys_platform != "darwin"
```
John-P, line 2: So this is for all platforms except macOS correct? I thought this was windows only.

shaneahmed: yes that's correct.

John-P (approved):

PR #473, opened 2022-09-16
BUG: Fix the `colorbar` Bug with `plt`

measty:
fixes the colorbar issue cropping up in test_visualization.py


shaneahmed (approved):
Thanks @measty

PR #476, opened 2022-09-23
ENH: Add Annotation Store Reader

measty:
This PR adds an AnnotationStoreReader class that allows an annotation store to be read as if it were a WSI. That can be used wherever a WSIreader can be used.

For example, to extract overlay patches around a few random glands using PointsPatchExtractor:
```
SQ = SQLiteStore(path\to\store.db)
glands = SQ.query(where='props["type"] == "Gland"')
centroids = np.array([[gland.geometry.centroid.x, gland.geometry.centroid.y] for gland in glands.values()])
#get 6 random centroids
rand_centroids = centroids[np.random.randint(0, len(centroids), 6),:]
ann_reader = AnnotationStoreReader(db_path, base_wsi_reader=wsi)
extractor = PointsPatchExtractor(ann_reader, rand_centroids, patch_size=(1024, 1024))

#show the 6 patches in a 2x3 grid
patch_im = np.zeros((2048, 3072, 3), dtype=np.uint8)
for i, patch in enumerate(extractor):
    patch_im[1024*(i//3):1024*(i//3 + 1), (i%3)*1024:(i%3+1)*1024, :] = patch
plt.imshow(patch_im)
plt.show()
```

pr_ex_3

Can be used to draw annotations over its base wsi as an 'annotated slide' as above, or just as a standalone overlay mask as below right.

only_mask_pr_ex

Also moves the annotation rendering code into the AnnotationRender class instead of in the AnnotationTileGenerator, so it can be reused as needed. It probably belongs more naturally there regardless.

Updates the rendering code a with a few further options, for example options for a blur to give a more heatmap-like effect:

 
blur_pr_ex

Or to allow rendering with outline only:

store_render_outline

https://tia-toolbox.readthedocs.io/en/add-store-reader/


John-P:
Please see comments on other PR:
- https://github.com/TissueImageAnalytics/tiatoolbox/pull/445#issuecomment-1271385492
- https://github.com/TissueImageAnalytics/tiatoolbox/pull/445#issuecomment-1271392213

tiatoolbox/annotation/storage.py
```diff
@@ -2082,6 +2082,7 @@ def _handle_pickle_callable_pquery(
         """
 
         def add_props_to_result(result, properties):
```
John-P, line 2084: Needs type annotations and docstring details.

John-P:
```suggestion
        def add_props_to_result(result: DefaultDict[str, set], propertiesDict[str, Any]) -> None:
```

tiatoolbox/tools/pyramid.py
```diff
@@ -144,6 +143,7 @@ def get_tile(
         level: int,
         x: int,
         y: int,
+        res: int = 1,
         pad_mode: str = "constant",
         interpolation: str = "optimise",
```
John-P, line 148: If we are standardizing on US spelling shall we use "optimize"?
```suggestion
        interpolation: str = "optimize",
```

tiatoolbox/tools/pyramid.py
```diff
@@ -481,6 +481,12 @@ def __init__(
         if renderer is None:
             renderer = AnnotationRenderer()
         self.renderer = renderer
+        self.overlap = int(1.5 * renderer.blur_radius)
```
John-P, line 484: Where does the magic number 1.5 come from?

John-P (commented):
Just a few minor comments. Looks good.

tiatoolbox/tools/pyramid.py
John-P: Can we make this a named constant or add a comment to explain why this number is here?

tiatoolbox/tools/pyramid.py
measty: It seems the most common usage is 'optimise'; for example optimise is used everywhere in wsireader.py, and pyramid.py. So maybe it would be best to leave it as is for the moment.

tiatoolbox/tools/pyramid.py
measty: added a comment to explain the 1.5

tests/test_wsireader.py
```diff
@@ -431,7 +443,6 @@ def test_find_optimal_level_and_downsample_openslide_interpolation_warning(
 
 def test_find_optimal_level_and_downsample_jp2_interpolation_warning(sample_jp2):
     """Test finding optimal level for mpp read with scale > 1.
-
```
shaneahmed, line 434: This should not be removed.

tests/test_wsireader.py
```diff
@@ -476,7 +487,6 @@ def test_find_optimal_level_and_downsample_power(sample_ndpi):
 
 def test_find_optimal_level_and_downsample_level(sample_ndpi):
     """Test finding optimal level for level read.
-
```
shaneahmed, line 479: same here.

tiatoolbox/wsicore/wsireader.py
```diff
@@ -18,10 +18,13 @@
 import tifffile
 import zarr
 from defusedxml import ElementTree
+from PIL import Image
 
 from tiatoolbox import utils
+from tiatoolbox.annotation.storage import AnnotationStore, SQLiteStore
```
shaneahmed, line 24: Did you get a chance to check for circular imports?

tiatoolbox/wsicore/wsireader.py
```diff
@@ -4444,3 +4451,509 @@ def read_bounds(
             )
 
         return im_region
+
+
+class AnnotationStoreReader(WSIReader):
```
shaneahmed, line 4456: Should it not be defined in annotation package?

tiatoolbox/wsicore/wsireader.py
measty: As far as I can see there are no circular imports

tiatoolbox/wsicore/wsireader.py
measty: As it is a subclass of WSIReader, I though the most reasonable place for it to be was in wsireader.py with all the other WSIreader subclassess

tests/test_wsireader.py
```diff
@@ -1848,7 +1882,6 @@ def test_tiled_tiff_tifffile(remote_sample):
     """Test fallback to tifffile for files which openslide cannot read.
 
     E.G. tiled tiffs with JPEG XL compression.
-
     """
```
shaneahmed, line 1885:
```suggestion

    """
```

shaneahmed (approved):

PR #454, opened 2022-08-30
[skip ci] DEV: Remove nbsphinx and replace it with myst-nb.

shaneahmed:
- Remove nbsphinx and replace it with myst-nb.
- Replace Example gallery with examples/README.md.
- Remove repetitive jupyter notebook cells from docs.
- Hide outputs related to progress bar.
- Remove nbsphinx from requirements.
- Add myst-nb in requirements.

https://tia-toolbox.readthedocs.io/en/dev-doc-use-myst-nb/

PR #479, opened 2022-09-25
Improve 09-multi-task-segmentation.ipynb

DavidBAEpstein:
This PR is ready for review.  The file 09-multi-task-segmentation.ipynb in branch doc-09 is the result of an overnight run on my Macbook, with ON_GPU=False. After the run, I changed this to the default ON_GPU=True. Would reviewers  please check on the final image in the output? The colour bar shows the background as black. However, I think the background looks grey, so I think this may be a bug, but I'm not sure. The colour bar is produced by tiatoolbox.visualization.overlay_prediction_mask, which has been worked on recently by @Srijay-lab and, perhaps, by @measty. 

Under Colab, there is CERTAINLY a bug, because the command "!pip install tiatoolbox" installs a version of tiatoolbox that is known to be buggy. I could change this to a command that installs a better version of tiatoolbox, but, the way things are currently set up, this would be changed back to the buggy version when it is merged into develop, so there is no point. Please look at my Issue#474.


measty:
I think the background looking gray is as expected, because the overlay is semi-transparent to allow the tissue underneath to also be seen. So the overlay is indeed black in the 'background' region, as indicated by the colorbar, but it appears gray in the image because it is overlaid on a more or less white background, which naturally ends up looking grayish.

shaneahmed:
@DavidBAEpstein The background is semi-transparent as @measty mentioned. This is the problem we face with dark mode on some websites. I have applied a fix in #454 using 
```mpl.rcParams["figure.facecolor"] = "white"  # To make sure text is visible in dark mode```

for matplotlib settings.

shaneahmed:
I have tested this notebook on Colab with ON_GPU=True. There were no issues. I will run this again after merging it in #454.

shaneahmed (approved):
Thanks @DavidBAEpstein It looks good to me.

I will test it on GPU in #454 and fix rendering with myst-nb.

shaneahmed:
@DavidBAEpstein I have now tested the notebook on Colab using GPU and it runs fine with pip install git+https://github.com/TissueImageAnalytics/tiatoolbox.git@develop

PR #483, opened 2022-10-05
DEV: Fix Circular import for wsireader

shaneahmed:
- Fix Circular import for wsireader


John-P:
Happy to accept this as a hotfix but I think that a circular import is a sign that there needs to be some restructuring.

PR #484, opened 2022-10-06
DEV: Fix Circular import `wsireader`

shaneahmed:
- Fix Circular import `wsireader`


John-P:
Happy to accept this as a hotfix if it avoids the issue but I think that a circular import is a sign that there needs to be some restructuring.

John-P (approved):

shaneahmed:
> Happy to accept this as a hotfix if it avoids the issue for active versions but I think that a circular import is a sign that there is an underlying issue and there needs to be some restructuring.

Agree

PR #489, opened 2022-10-14
BUG: Read TIFF Resolution Tags As A Fraction

John-P:
This PR fixes a bug where only the numerator of the TIFF resolution tags was being read.


shaneahmed (approved):
Looks good.

Thanks @John-P

PR #492, opened 2022-10-20
BUG: Fix Bug When OME-TIFF Missing Objective Power

John-P:
This PR refactors the code fro parsing OME-TIFF metadata and fixed a bug where an exception would be raised if the OME XML is missing objective power. This was noticable when there was MPP data but not objective power such as with file created by tifffile.

### DIFF
--- a/tiatoolbox/wsicore/wsireader.py
+++ b/tiatoolbox/wsicore/wsireader.py
@@ -10,13 +10,14 @@ import re
 import warnings
 from datetime import datetime
 from numbers import Number
-from typing import Iterable, Optional, Tuple, Union
+from typing import Iterable, List, Optional, Tuple, Union
 
 import numpy as np
 import openslide
 import pandas as pd
 import tifffile
 import zarr
+from defusedxml import ElementTree
 
 from tiatoolbox import utils
 from tiatoolbox.utils.env_detection import pixman_warning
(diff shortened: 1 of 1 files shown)


tiatoolbox/wsicore/wsireader.py
```diff
@@ -3167,36 +3168,69 @@ def time(string: str) -> datetime:
...
+            dict:
+                Dictionary of kwargs for WSIMeta.
+
+        """
         # The OME-XML should be in each IFD but is optional. It must be
         # present in the first IFD. We simply get the description from
         # the first IFD.
-        from defusedxml.ElementTree import fromstring as et_from_string
-
         description = self.tiff.pages[0].description
```
shaneahmed, line 3193: you are using this line twice once here and then in `_get_ome_xml`. you can get this from `_get_ome_xml`

shaneahmed (commented):
Just one minor comment.
Please can you also check coverage.

tiatoolbox/wsicore/wsireader.py
John-P: This line is being removed. It should only be used on one line in `_get_ome_xml`.

shaneahmed: how about this 
https://github.com/TissueImageAnalytics/tiatoolbox/blob/6d714f4d2dcf5bed053d0f08f0792176b2a5cd60/tiatoolbox/wsicore/wsireader.py#L3193

shaneahmed (approved):
Thanks @John-P

PR #498, opened 2022-10-22
DEV: Update Dependencies

shaneahmed:
- Update dependencies

https://tia-toolbox.readthedocs.io/en/dev-update-dependencies/


docs/requirements.txt
```diff
@@ -1,8 +1,7 @@
-docutils>=0.17.1
 furo>=2022.6.21
-jinja2<3.1.0
-myst-nb>=0.16.0
-myst-parser>=0.17.0
+jinja2
```
shaneahmed, line 2: @John-P please can you confirm why it was pinned to less than 3.1.0?

John-P: This was pinned because of an incompatibility with another library. I can't remember which but it should have been included in a comment somewhere.

John-P: I believe it was the current version of flask to work. This may have been fixed on the latest flaks but I have not tested this.

shaneahmed: ok I will leave this PR for now.

requirements.txt
```diff
@@ -1,31 +1,31 @@
...
+SimpleITK>=2.1.1
+sphinx>=5.3.0
 tifffile>=2021.11.2
-torch>=1.11.0
-torchvision>=0.12.0
-tqdm>=4.63.1
-umap-learn>=0.5.2
-wsidicom>=0.2.0
-zarr>=2.11.1
+torch>=1.12.1
```
shaneahmed, line 26:
```suggestion
torch>=1.13.0
```

docs/requirements.txt
John-P: Looks like it might be fixed on sphinx. I see their requirements has >=3.0: https://github.com/sphinx-doc/sphinx/blob/cd3f2e435000835dd98a11497440bc16a79ec31c/pyproject.toml#L64

Also, readthedocs also automatically installs [sphinx-jinja2-compat](https://pypi.org/project/sphinx-jinja2-compat/) during our builds.

It must be an issue for local builds still. Did you get it to build on your setup with new jinja2 (>3.1)?

shaneahmed: Local build also installs [sphinx-jinja2-compat](https://pypi.org/project/sphinx-jinja2-compat/) and ignores latest version of jinja2 in favour of 3.0.3.

PR #503, opened 2022-11-02
DOCS: Simplify Example Notebook 1

John-P:
This PR aims to simplify the first notebook to make it more readable and very simple for first time users. It is now shorter, with less scary looking code (e.g. complex code for just downloading an example image) and easier for new people to follow.
Docs: https://tia-toolbox.readthedocs.io/en/docs-examples-1/_notebooks/jnb/01-wsi-reading.html

## Changes

- Add a function to the data module to download a simple test image for examples, similar to scikit-image (See below).
- Change masking section to use the `WSIReader.tissue_mask` function and return as `VirtualWSIReader`.
- Remove section of SLIC superpixels.

```python
import tiatoolbox

sample_file_path = tiatoolbox.data.small_svs()
```


shaneahmed (approved):
Looks Good to me.

PR #504, opened 2022-11-03
🐳 DEV: Publishing Docker Images 

mostafajahanifar:
This PR introduces TIAToolbox docker images and a workflow to publish them as packages on `ghcr.io`.

Github Container Registry (`ghcr.io`) is similar to `docker.io` for hosting toolbox packages and allows using Github actions to automatically publish packages on schedule (for example, on new releases).

However, because releasing TIAToolbox packages is never straightforward (requires testing, changing versions and OS depending on requirements, etc.) and it is important not to publish bad containers (because public containers cannot be deleted), I decided to make a workflow action which is manually triggered but builds and publishes all the packages automatically.

There are two important parts to this PR:
1- Docker images introduced in `./docker/` directory which are separated based on different Python and running OS (base image).
2- The workflow action introduced in `./.github/workflows/docker-publish.yml` which manages the build and publishing.

Following [George's suggestion](https://github.com/TissueImageAnalytics/tiatoolbox-docker/issues/2), the publish action automatically tags the containers using `<toolbox_version>-py<python_version>-<os_version>` convention to make it easier for users to understand what is included in each image. For example, someone would be able to download TIAToolbox 1.3.0 with python 3.9 on Ubuntu by specifying this base image in their work:
```
FROM ghcr.io/tissueimageanalytics/tiatoolbox:1.3.0-py3.9-ubuntu
```
We also generate a simple `tiatoolbox:latest` tag based on the latest Ubuntu OS and Python version.

The `docker-publish.yml` workflow automatically manages tagging, provided with the right `TOOLBOX_VER`. As docker images should be checked on each release and only repo owner can publish to `ghcr.io/tissueimageanalytics/tiatoolbox`, this action can only be called through `develop` branch and by Shan (or whoever has admin access). Therefore, on each release, a new PR should be created to make appropriate changes to `docker-publish.yml` (changing the version tag or adding/removing different verison) and docker images (if needed) and then merge that PR into develop to be able to trigger the new action.


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,60 @@
+name: Publish cpu dockers
+on:
+  workflow_dispatch
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: TissueImageAnalytics/tiatoolbox
+  image: ghcr.io/tissueimageanalytics/tiatoolbox
+  TOOLBOX_VER: 1.3.0
```
shaneahmed, line 9: What will happen when we update the version?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,60 @@
+name: Publish cpu dockers
```
shaneahmed, line 1:
```suggestion
name: Publish CPU dockers
```

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,60 @@
...
+          - dockerfile: ./3.7/Debian/Dockerfile
+            mtag: py3.7-debian
+          - dockerfile: ./3.8/Debian/Dockerfile
+            mtag: py3.8-debian
+          - dockerfile: ./3.9/Debian/Dockerfile
+            mtag: py3.9-debian
+          - dockerfile: ./3.9/Ubuntu/Dockerfile
+            mtag: py3.9-ubuntu
+          - dockerfile: ./3.9/Ubuntu/Dockerfile
+            mtag: latest
```
shaneahmed, line 27: How about python 3.10?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,60 @@
+name: Publish cpu dockers
+on:
+  workflow_dispatch
```
shaneahmed, line 3: Trigger on Tags?

.github/workflows/docker-publish.yml
John-P: @shaneahmed This should 🤞 be updated by bump2version. But we may need to make sure this file is in the paths which is checks for version strings.

.github/workflows/docker-publish.yml
John-P: Good suggestion. We could have a special docker release tag if we don't want the regular release tag to trigger this for some reason.

docker/3.9/Ubuntu/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN apt-get update && apt-get install --no-install-recommends -y python3.9 python3.9-dev python3.9-venv python3-pip python3-wheel build-essential && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# create and activate virtual environment
+RUN python3.9 -m venv /opt/venv
+ENV PATH="/opt/venv/bin:$PATH"
+
+RUN pip3 install --no-cache-dir tiatoolbox
+
+FROM ubuntu:20.04 AS runner-image
```
shaneahmed, line 15:
```suggestion
FROM ubuntu:22.04 AS runner-image?
```

.github/workflows/docker-publish.yml
mostafajahanifar: As we discussed in the meeting, I'm against automatic triggering of this as it would be useless. For each publish, image files should be checked if they are working, dependencies should be checked, newer version of OS should be added or unsupported version should be removed. So, instead of automatic trigger, for each publish we should have a separate PR that checks for these and makes the necessary updates to the files and then after merging that PR with develop, this can be triggered from develop branch actions. 

We also we want to make sure that the release is completely successful (no hot bugs) before updating the docker image registry because we don't want to break any application that uses those packages.

I'm not against adding automatic Tags based on version strings tho. However, if we are going to check everything manually before publishing, I don't see a point for going that way either.

.github/workflows/docker-publish.yml
mostafajahanifar: Sure, I will add it.

.github/workflows/docker-publish.yml
mostafajahanifar: Please check my comment here: https://github.com/TissueImageAnalytics/tiatoolbox/pull/504#discussion_r1018959664


I'm not sure how necessary is this and I don't know how this can be realized.

.github/workflows/docker-publish.yml
shaneahmed: As discussed we will trigger this manually.

.github/workflows/docker-publish.yml
John-P: What about D for docker? :p

docker/3.9/Ubuntu/Dockerfile
mostafajahanifar: Done in [7f030d8](https://github.com/TissueImageAnalytics/tiatoolbox/pull/504/commits/7f030d821f83514aad5385703be701e856f583a4)

.github/workflows/docker-publish.yml
mostafajahanifar: Done in [7f030d8](https://github.com/TissueImageAnalytics/tiatoolbox/pull/504/commits/7f030d821f83514aad5385703be701e856f583a4)

shaneahmed (approved):

PR #511, opened 2022-11-22
DEV: Update bump2version config

shaneahmed:
- Update bump2version config for citation and docker

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,6 +11,14 @@ replace = version="{new_version}"
 search = __version__ = "{current_version}"
 replace = __version__ = "{new_version}"
 
+[bumpversion:file:CITATION.cff]
+search = version: "{current_version}"  # TIAToolbox version
+replace = version: "{new_version}"  # TIAToolbox version
+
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -22,7 +30,7 @@ spellcheck-targets = comments
 dictionaries = en_US,python,technical
 max-cognitive-complexity = 14
 max-expression-complexity = 7
-per-file-ignores = 
+per-file-ignores =
 	test_*: E800
```
John-P, line 34: Why are we ignoring commented out code in test files?

John-P (commented):
Happy with the change to bump2version config but unsure about ignore E800

setup.cfg
shaneahmed: I think it should flagged on individual lines in the code. Let me create a separate PR for this as I expect there would be errors after removing this.

John-P (approved):
Looks good. Happy to review flake8 E800 exception in a another PR.

PR #512, opened 2022-11-24
DEV: Fix Flake8 E800 for tests

shaneahmed:
- Update `setup.cfg` for flake8 E800.
- Enables `E800` checks - "Found commented out code".
- Remove DeepSource ignore flags from test files.
- Update `requirements.txt` files
- Move comment in `setup.cfg` which is now causing an error.
- Fix DeepSource Errors.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
 furo>=2022.6.21
-jinja2>=3.0.3
+jinja2>=3.0.3, <3.1.0
 myst-nb>=0.17.1
 myst-parser>=0.18.1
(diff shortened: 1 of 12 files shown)


John-P:
The problem is https://github.com/TissueImageAnalytics/tiatoolbox/blob/3272bbfd17ae27fb77cf7122cf20cd95762e0896/setup.cfg#L28

Flake8 is trying to parse the comments and is hitting the '#' which is causing an error.
Maybe try moving the comment to the line before or after?

John-P (approved):

PR #513, opened 2022-11-25
🚨  Expand Tile Server API

measty:
This PR adds to the tileserver a variety of endpoints required by the visualization interface for changing slide, changing overlays, and controlling various aspects of how rendered annotations are displayed. Use these endpoints to control the tiles served by the tileserver.

added routes are:

```
GET: "/tileserver/session_id": setup a session id upon a new connection
PUT: "/tileserver/color_prop": change the property used to determine the color of rendered annotations
PUT: "/tileserver/slide":  change the slide as specified in the path
PUT: "/tileserver/cmap": change colourmap
PUT: "/tileserver/annotations": load some annotations, adding to existing if present
PUT: "/tileserver/overlay": change or add an overlay. If provided a path to a file containing annotations, current annotation overlay layer is replaced. Image-based overlays are added as new layers.
POST: "/tileserver/commit": save the current overlay database
PUT: "/tileserver/renderer/<prop>": generic updater for renderer props that don't need specific endpoint to handle them
PUT: "/tileserver/secondary_cmap": specify a colormap override for one specific type
GET: "/tileserver/prop_names/<ann_type>": get all unique properties that appear on annotations in the store
GET: "/tileserver/prop_values/<prop>/<ann_type>": get all values of property over annotations of specified type (or all)
GET: "/tileserver/reset/<session_id>": reset the server, removing info associated with session
GET: "/tileserver/color_prop": get the property used to determine the color of rendered annotations
GET: "/tileserver/slide":  get the current slide info
GET: "/tileserver/cmap": get colourmap
GET: "/tileserver/annotations": get annotations in a within given geometry
GET: "/tileserver/overlay": get current annotation overlay
GET: "/tileserver/renderer": get a renderer prop that doesn't need specific endpoint to handle it
GET: "/tileserver/secondary_cmap": get the colormap override for one specific type
```

Further details of the route specifics and required parameters can be found in the tileserver_api.yml specification.

I've updated it to use more appropriate verbs and have added an openAPI .yml file to describe the api.

The main added features are:

1. Provides an endpoint to change the base slide that the tileserver serves tiles from. Paths provided to the tileserver in this endpoint and others should be in url encoded form (for example the result of urllib.parse.quote(path)).
2. Provides an endpoint to add an overlay to the tileserver from a path to an overlayable source. This can be either a pyramid image of the same size as the base slide, or a low resolution non-pyramidal overlay image of the same aspect ratio, or a source of annotations in a compatible format such as a tiatoolbox SQLiteStore .db file, or a .geojson file. The overlay will be added as a layer. The tileserver can support one annotation-based overlay, and an unlimited number of image-based overlays. An annotation overlay will be added with layer name 'overlay', replacing any previous annotation layer. An image overlay will be added as 'layer-N' where N is the number of image overlays present including the newly added one.
3. Provides an endpoint to load new annotations from a file. These will be added to the current annotation overlay in a non-permanent way. To save added annotations to the .db (or save a .db of the current annotations if it was loaded from another source such as .geojson), a 'commit' endpoint is provided.
4. Provides a variety of endpoints to control how the annotations are rendered when generating the annotation overlay layer. This includes operations such as modifying the property used to determine the colour of rendered annotations, the ability to specify the colourmap used to calculate the colour from the property, and the ability to set various renderer properties such as edge thickness, filters on which annotations are rendered, etc. See AnnotationRenderer class documentation for details of properties and their effects.


John-P:
Rather than using verbs in the URL routes I would suggest using proper [HTTP verbs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) such as GET and POST and to turn this into a proper [REST API](https://www.redhat.com/en/topics/api/what-is-a-rest-api).

John-P:
I would also suggest making an OpenAPI YAML file to describe these endpoints. E.g. I am unclear what /setup does at the moment.

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,290 @@
...
+                    type: array
+                    items:
+                      type: string
+
+  /tileserver/reset:
+    get:
+      summary: Remove no longer needed objects if user closed the browser
+      responses:
+        '200':
+          description: Successful response
```
John-P, line 290: Is this action on some kind of timeout? E.g. if the user loses connection or the browser crashes? Else the clean up will never be called.

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,290 @@
...
+          schema:
+            type: string
+      responses:
+        '200':
+          description: jsonified list of types present in the overlay, if any
+          content:
+            application/json:
+
+  /tileserver/commit/{save_path}:
+    get:
```
John-P, line 176:
```suggestion
    post:
```

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,290 @@
...
+          name: save_path
+          description: Url safe path to save the database changes
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Successful response
+
+  /tileserver/update_renderer/{prop}/{val}:
```
John-P, line 189:
```suggestion
  /tileserver/renderer/{prop}/{val}:
```

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,290 @@
...
+          name: val
+          description: New value for the property
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Successful response
+
+  /tileserver/change_secondary_cmap/{type_id}/{prop}/{cmap}:
```
John-P, line 209:
```suggestion
  /tileserver/secondary_cmap/{type_id}/{prop}/{cmap}:
```

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,290 @@
...
+          name: cmap
+          description: Color map to use for the property
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Successful response
+
+  /tileserver/get_prop_names/{ann_type}:
```
John-P, line 236:
```suggestion
  /tileserver/prop_names/{ann_type}:
```

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,290 @@
...
+            type: string
+      responses:
+        '200':
+          description: List of annotation properties.
+          content:
+            application/json:
+              schema:
+                type: array
+
+  /tileserver/get_prop_values/{prop}/{ann_type}:
```
John-P, line 255:
```suggestion
  /tileserver/prop_values/{prop}/{ann_type}:
```

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,290 @@
...
+              schema:
+                type: object
+                properties:
+                  values:
+                    type: array
+                    items:
+                      type: string
+
+  /tileserver/reset:
+    get:
```
John-P, line 286:
```suggestion
    put:
```

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,290 @@
...
+            or a dict string
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Successful response
+
+  /tileserver/load_annotations/{file_path}/{model_mpp}:
+    put:
```
John-P, line 134:
```suggestion
    post:
```

If this is adding a new thing I would use POST and for updating existing use PUT.

John-P (changes requested):
Nice expansion of the capabilities of the tile server.

Some suggestions on the API are to remove verbs from the route and change some of the HTTP verbs for some routes.

I am also concerned about user authentication and management. That is a massive can or worms (many security issues etc) and think we should not be managing that. I suggest leaving it to someone else such as oauth etc. Else it should be outside the scope of the toolbox.

tiatoolbox/visualization/tileserver_api.yml
measty: Bokeh has some hooks you can use to do stuff when specific things happen. So in the bokeh app, I have this set up to be called when a bokeh session is destroyed (eg if they close the window)

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,290 @@
...
+          description: Successful response
+          headers:
+          Set-Cookie:
+            schema:
+              type: string
+            description: The 'user' cookie containing the user identifier.
+
+  /tileserver/change_color_prop/{prop}:
+    put:
+      summary: Update property to color annotations by
```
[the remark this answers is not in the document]
measty: it isnt missing anything - but perhaps I could phrase it better. It updates the property that is used to determine the annotations rendered color (for example we might color by type, or color by BAM score, or whatever)

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,290 @@
...
+      description: This endpoint serves a web page with a simple slide viewer
+      responses:
+        '200':
+          description: The index page.
+          content:
+            text/html:
+              schema:
+                type: string
+
+  /tileserver/get_user:
```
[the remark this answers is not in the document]
measty: this endpoint just returns a randomly generated ID, that will be used to identify requests coming from a particular session to disambiguate them from requests coming from another session

measty:
I've updated the routes as requested so that they dont include the verbs, and have also renamed all the 'user' related stuff to 'session' to make clear its a temporary session id and not a persistent user login type thing.

I think it should be ready for merging now

tests/test_tileserver.py
```diff
@@ -1,21 +1,37 @@
...
+from matplotlib import colormaps
 from shapely.geometry import LineString, Polygon
 from shapely.geometry.point import Point
 
 from tests.test_annotation_stores import cell_polygon
+from tests.test_utils import make_simple_dat
 from tiatoolbox.annotation.storage import Annotation, AnnotationStore, SQLiteStore
 from tiatoolbox.cli.common import cli_name
-from tiatoolbox.utils.misc import imwrite
+from tiatoolbox.utils.misc import imread, imwrite
```
shaneahmed, line 19: You can merge #606 and import as below
```suggestion
from tiatoolbox.utils import imread, imwrite
```

tests/test_tileserver.py
```diff
@@ -1,21 +1,37 @@
...
 from tiatoolbox.visualization.tileserver import TileServer
 from tiatoolbox.wsicore.wsireader import WSIReader
```
shaneahmed, line 21:
```suggestion
from tiatoolbox.wsicore import WSIReader
```

tiatoolbox/visualization/tileserver.py
```diff
@@ -63,33 +70,110 @@ def __init__(
...
+        if cmap[0] == "{":
+            cmap = ast.literal_eval(cmap)
+
+        if cmap == "None":
+            return None
+        if isinstance(cmap, str):
+            return colormaps[cmap]
+
+        def cmapp(x):
+            """Dictionary colormap callable wrapper"""
```
shaneahmed, line 173:
```suggestion
            """Dictionary colormap callable wrapper."""
```

tiatoolbox/visualization/tileserver.py
```diff
@@ -160,18 +263,46 @@ def zoomify(
 
         """
         try:
-            pyramid = self.tia_pyramids[layer]
+            pyramid = self.tia_pyramids[session_id][layer]
```
shaneahmed, line 266: Why we name these as "tia_pyramids"?

shaneahmed:
> This draft PR adds to the tileserver a variety of endpoints required by the visualization interface for changing slide, changing overlays, and controlling various aspects of how rendered annotations are displayed. Use these endpoints to control the tiles served by the tileserver.
> 
> added routes are:
> 
> ```
> GET: "/tileserver/session_id": setup a session id upon a new connection
> PUT: "/tileserver/color_prop/<prop>": change the property used to determine the color of rendered annotations
> PUT: "/tileserver/slide/<slide_path>":  change the slide as specified in the path
> PUT: "/tileserver/cmap/<cmap>": change colourmap
> PUT: "/tileserver/load_annotations/<file_path>/<float:model_mpp>": load some annotations, adding to existing if present
> PUT: "/tileserver/overlay/<overlay_path>": change or add an overlay. If provided a path to a file containing annotations, current annotation overlay layer is replaced. Image-based overlays are added as new layers.
> GET: "/tileserver/commit/<save_path>": save the current overlay database
> PUT: "/tileserver/renderer/<prop>/<val>": generic updater for renderer props that don't need specific endpoint to handle them
> PUT: "/tileserver/secondary_cmap/<type_id>/<prop>/<cmap>": specify a colormap override for one specific type
> GET: "/tileserver/prop_names/<ann_type>": get all unique properties that appear on annotations in the store
> GET: "/tileserver/prop_values/<prop>/<ann_type>": get all values of property over annotations of specified type (or all)
> GET: "/tileserver/reset/<session_id>": reset the server, removing info associated with session
> ```
> 
> I've updated it to use more appropriate verbs and have added an openAPI .yml file to describe the api.

@measty Please can you update this description. If you can highlight the main features added that would be great.

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,290 @@
...
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Successful response
+
+  /tileserver/load_annotations/{file_path}/{model_mpp}:
+    put:
+      summary: Load annotations for a slide
```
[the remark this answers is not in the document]
measty: This endpoint adds annotations to the current annotation overlay layer, if there is one existing already. If one doesn't exist, then one is created. I've assigned it as 'put', because it is putting additional annotations into the annotation overlay layer (by loading them from a file).

measty:
I've updated the description to highlight the added functionality, removed the tia_ from some variable names, and changed the imports as suggested.

John-P:
The URLs still don't look right to me. There should be corresponding PUT and GET requests should be for the same resource and PUT parameters shouldn't be in the URL unless it is specifying the resource itself. E.g. PUT: "/tileserver/cmap/<cmap>" shouldn't have <cmap> in the URL but in the payload. This is because GET: "/tileserver/cmap/" should return the current cmap and PUT: "/tileserver/cmap/" should set it. See https://www.rfc-editor.org/rfc/rfc7231#section-4.3.4.

measty:
> The URLs still don't look right to me. There should be corresponding PUT and GET requests should be for the same resource and PUT parameters shouldn't be in the URL unless it is specifying the resource itself. E.g. PUT: "/tileserver/cmap/" shouldn't have `<cmap>` in the URL but in the payload. This is because GET: "/tileserver/cmap/" should return the current cmap and PUT: "/tileserver/cmap/" should set it. See https://www.rfc-editor.org/rfc/rfc7231#section-4.3.4.

@John-P I've put all the parameters for the 'put' routes in the payload, added 'get' requests for the 'put' requests that didn't already have them, and changed the tests/yaml accordingly.

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,399 @@
...
+      requestBody:
+        required: true
+        content:
+          application/x-www-form-urlencoded:
+            schema:
+              type: object
+              properties:
+                cmap:
+                  type: string
+                  description: Name of the colormap, should be a matplotlib colormap or a dict string
```
John-P, line 154: @measty should this be 'urlencoded' if it's not in the URL?

John-P: I'm not sure familiar with the openapi yaml schema so I may be wrong but it just didn't look quite right.

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -0,0 +1,399 @@
...
+                      type: object
+
+  /tileserver/overlay:
+    put:
+      summary: Change overlay for a slide
+      description: This endpoint changes the overlay. If the path points to some annotations, the current overlay is replaced with the new one. If the path points to an image, it is added as a new layer.
+      requestBody:
+        required: true
+        content:
+          application/x-www-form-urlencoded:
```
John-P, line 226: Same thing here, is this right?

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -99,7 +99,7 @@ paths:
               properties:
                 prop:
                   type: string
-                  description: The property to color annotations by
+                  description: json encoded property name to color annotations by
```
John-P, line 102: Minor suggestion: json -> JSON

tiatoolbox/visualization/tileserver_api.yml
```diff
@@ -197,10 +197,10 @@ paths:
               properties:
                 bounds:
                   type: list
-                  description: a list of 4 numbers, [xmin, ymin, xmax, ymax] describing the bounds of the query
+                  description: a json encoded list of 4 numbers, [xmin, ymin, xmax, ymax] describing the bounds of the query
                 where:
                   type: string
-                  description: a valid AnnotationStore filter string
+                  description: a valid AnnotationStore filter string, json encoded
```
John-P, line 203: a -> A

measty: I've corrected the capitalization in the instances you pointed out and a few other ones too

PR #517, opened 2022-12-02
DOC: Copy Sample SVS Loader From docs-examples-1

John-P:
This mirrors a subset of changes from #503 to add a sample SVS loading function `tiatoolbox.data.small_svs()` to the data module for use in documentation examples.


shaneahmed (approved):
Thanks @John-P

Issue #532, opened 2023-02-02
Documentation inconistency


blaginin:
- TIA Toolbox version: 1.3.1, dev
- Python version: 3.10.8 (conda)
- Operating System: MacOS 13.1 (22C65)

### Description

Trying to learn the library and discovered some cases where docs are inconsistent with the current behaviour. 

### What I Did

- [x] Tried to create `AnnotationStore` (as it was stated [here](https://github.com/TissueImageAnalytics/tiatoolbox/blob/76fcfff0593e4dbb55c0bc58b687d3788cdd33e3/tiatoolbox/annotation/storage.py#L833), [here](https://github.com/TissueImageAnalytics/tiatoolbox/blob/76fcfff0593e4dbb55c0bc58b687d3788cdd33e3/tiatoolbox/annotation/storage.py#L920), [here](https://github.com/TissueImageAnalytics/tiatoolbox/blob/76fcfff0593e4dbb55c0bc58b687d3788cdd33e3/tiatoolbox/annotation/storage.py#L931), [here](https://github.com/TissueImageAnalytics/tiatoolbox/blob/76fcfff0593e4dbb55c0bc58b687d3788cdd33e3/tiatoolbox/annotation/storage.py#L2153), [here](https://github.com/TissueImageAnalytics/tiatoolbox/blob/76fcfff0593e4dbb55c0bc58b687d3788cdd33e3/tiatoolbox/annotation/storage.py#L2408), and [here](https://github.com/TissueImageAnalytics/tiatoolbox/blob/76fcfff0593e4dbb55c0bc58b687d3788cdd33e3/tiatoolbox/annotation/storage.py#L2408)).
```python
>>>  store = AnnotationStore()
TypeError: Can't instantiate abstract class AnnotationStore with abstract methods __getitem__, __len__, __setitem__, commit, dump, dumps, open
```
`AnnotationStore` has become abstract, and it would be helpful to update the docs accordingly. Examples can either state that users must use derived classes or explicitly show methods usage on `SQLiteStore` or `DictionaryStore`.

[...]

PR #540, opened 2023-02-16
:sparkles: ENH: Add Neighbourhood Querying Support To `AnnotationStore`

John-P:
Enable easy and efficient querying of annotations within a neighbourhood of other annotations. This is planned to be executed in the query domain (e.g. by sqlite, outside of the Python GIL) where possible which should be faster than a user writing a for loop and performing many queries for this common use case.

## Docs

<a href="https://tia-toolbox.readthedocs.io/en/feature-neighbourhood-query/_autosummary/tiatoolbox.annotation.storage.AnnotationStore.html#tiatoolbox.annotation.storage.AnnotationStore.nquery"></a>

## Initial Concept

Pseudo query example

```
SELECT purple WITHIN 10 points of red
```

Neighbourhood Querying drawio

This is now implemented as:

```python
results = store.nquery(
  where="props['color'] == 'red'",
  n_where="props['color'] == 'purple'",
  distance=10,
  # Use centre points of bounding boxes for distance
  mode="boxpoint-boxpoint",
)
```

### Query Modes

<table style="td {text-align: center;}">
<tr>
<td>
poly-poly
</td>
<td>
box-box
</td>
</tr>
</tr>
<td>
boxpoint-boxpoint
</td>
</tr>
</table>

## In this PR

- Fix special "bbox_intersects" for `DictionaryStore`. This special geometry predicate was added as a special optimized case which is true if the bounding box of the query geometry intersects the bounding box of the stored geometry. This was never implemented for `DicstionaryStore`.
- Add geometry predicate "bbox_centre_within" which functions similarly to the special "bbox_intersects" predicate. "bbox_centre_within" returns True if the centre of the bounding box of the stored geometry is within the query bounds.
- Add `nquery` function which allows for querying for annotations which are within the neighbourhood of other annotations. This is done by supplying a query geometry and a where predicate to select the initial annotations to use as the centre of each neighbourhood. A second `n_where` predicate, `distance`, and `mode` argument are used to find neighbours.
  - `distance` is the size of the neighbourhood to search (see also `mode`). For a polygon-polygon check, this is via a buffer applied to the geometry, for point-point checks this is a radius, for box-box checks this is added to each side of the bounding box.
  - `n_where` is a predicate, similar to `where`, but used for filtering neighbours after selecting the annotations to query around via `where`.
  - `mode` may be "poly-poly" (full expensive polygon-polygon intersection), "box-box" (bounding box intersection) or "boxpoint-boxpoint" (centre of bounding box within `distance`). Currently, supported modes are only "poly-poly", "box-box", and "boxpoint-boxpoint". Others are possible, but would complicate implementation of these two sets. This would lose information about exactly which annotation was within the neighbourhood of another, but would be more efficient at finding intersections of two groups. Perhaps this could simply be shown as a 'recipe' in the documentation.

## Benchmarking

Some initial benchmarking shows that `SQLiteStore` significantly outperforms `DictionaryStore`.

In this test an $n \times n$ grid of cell boundaries is generated and overlaid with another $n \times n$ grid. A query is then performed to find overlapping geometries via either any overlap of the bounding boxes (box-box) or the centre of the bounding boxes (boxpoint-boxpoint) within a distance $k$. Therefore, the largest test here where $n=100$ is a grid of $100 \times 100$ artificial cell boundaries labelled as class "A", overlaid with another $100 \times 100$ grid of cell boundaries labelled with class "B" for a total of $20,000$ geometries.

This plot is produced from runs on a 6 core Intel(R) Core(TM) i5-8500 CPU @ 3.00GHz CPU:

nquery2

It is curious that the "boxpoint-boxpoint" query mode is slower than "poly-poly". I have a feeling that it may have something to do with Shapely 2.0 optimisations where it can vectorise batch geometry operations, or maybe there is some condition for the polygon-polygon intersection that allows for fast failing of tests.

## Notes

- It is already much faster for SQLiteStore but this could be likely improved further by using a subquery.
- The implementation currently performs one query to get all annotations which will be the centre of $m$ secondary queries performed via a loop in Python. This could be refactored to be a nested sqlite subquery which will likely be faster. I think this optimisation could be a new PR though as this basic functionality is already usable.
- This is currently implemented in the base class, so all subclasses get it for 'free'.
- Another approach which may be faster for some cases (but more limited) is to query for initial annotations, then apply a distance buffer, merge overlapping geometries (reduce total number assuming many overlaps), query for all candidate matches once, compute intesection.

## To-Do

- [x] Test coverage (it's as good as I can get it for now)
- [x] Fix linter issues
- [x] Add diagrams to docstrings?
- [x] Add support for um units as distance? (Decided to do in a follow up PR)
- [x] Improve docstrings and add more examples? (suggestions welcome)
- [x] More tests? (suggestions welcome)


shaneahmed:
Please can you also fix documentation issue as highlighted here #532

benchmarks/annotation_nquery.py
```diff
@@ -0,0 +1,196 @@
...
+    repeat_first: bool = True,
+    direction: str = "CCW",
+) -> Polygon:
+    """Generate a fake cell boundary polygon.
+
+    Cell boundaries are generated an ellipsoids with randomised eccentricity,
+    added noise, and a random rotation.
+
+    Args:
+        xy (tuple(int)): The x,y centre point to generate the cell boundary around.
```
shaneahmed, line 34:
```suggestion
        xy (tuple(int)): 
            The x,y centre point to generate the cell boundary around.
```
Description on next line with an indent just to be consistent with the rest of the repo as you did before?

benchmarks/annotation_nquery.py
```diff
@@ -0,0 +1,196 @@
...
+        n_points (int): Number of points in the boundary. Defaults to 20.
```
shaneahmed, line 35:
```suggestion
        n_points (int): 
            Number of points in the boundary. Defaults to 20.
```

benchmarks/annotation_nquery.py
```diff
@@ -0,0 +1,196 @@
...
+)
+
+
+# Copied from tests, refactor into a common location?
+def cell_polygon(
+    xy: Tuple[Number, Number],
+    n_points: int = 20,
+    radius: Number = 10,
+    noise: Number = 0.01,
+    eccentricity: Tuple[Number, Number] = (1, 3),
```
shaneahmed, line 24: Number is not consistent with float as you mention in the docstring.

benchmarks/annotation_nquery.py
shaneahmed, line 1: Wouldn't it be better if this is written as jupyter notebook? As you can then upload it as a document which anyone can read with the resutls online without running it?

tests/test_annotation_stores.py
```diff
@@ -1642,6 +1650,437 @@ def test_connection_to_path_io(store_cls, tmp_path):
...
+            where="props['class'] == 'A'",
+            n_where="props['class'] == 'B'",
+            distance=distance,
+            mode="boxpoint-boxpoint",
+        )
+        assert isinstance(result, dict)
+        assert len(result) == 1
+
+    @staticmethod
+    def test_nquery_overlapping_grid_box_box(store_cls):
```
shaneahmed, line 1950: Docstring missing?

benchmarks/annotation_nquery.py
John-P: Personally no, I think it's better as a script as it takes several hours to run.

John-P: I think benchmarks generally are better as scripts not notebooks.

John-P: I'm also done with this and not going to re-write it!

John-P: Done it now!

tests/test_dsl.py
```diff
@@ -91,6 +91,21 @@ def sqlite_eval(query: Union[str, Number]):
     return result
 
 
+class TestSQLite:  # noqa: PIE798
+    """Test converting from our DSL to an SQLite backed."""
+
```
measty, line 96: should be "backend"

tiatoolbox/annotation/storage.py
```diff
@@ -1007,6 +1050,195 @@ def select_values(
...
+        bounding box of stored annotations and the query geometry when
+        determining if annotations are within the neighbourhood.
+
+        .. figure:: ../images/nquery-boxpoint-boxpoint.png
+            :width: 512
+            :alt: "boxpoint-boxpoint" mode
+
+
+        Examples:
+            Example bounding boxy query with one neighbour within a
```
measty, line 1151: Example bounding box query

tiatoolbox/annotation/storage.py
```diff
@@ -1007,6 +1050,195 @@ def select_values(
...
+            distance of 2.0.
+
+            >>> from shapely.geometry import Point, Polyon
+            >>> from tiatoolbox.annotation.storage import Annotation, SQLiteStore
+            >>> store = SQLiteStore()
+            >>> annotation = Annotation(Point(0, 0), {"class": 42})
+            >>> store.add(annotation, "foo")
```
measty, line 1158: Shouldnt it be append() not add ()?

tiatoolbox/annotation/storage.py
```diff
@@ -1007,6 +1050,195 @@ def select_values(
...
+        distance: float = 5.0,
+        geometry_predicate: str = "intersects",
+        mode: str = "poly-poly",
+    ) -> Dict[str, Dict[str, Annotation]]:
+        """Query for annotations within a distance of another annotation.
+
+        Args:
+            geometry (Geometry):
+                A geometry to use for the query. If None, the geometry of
+                the annotation is used. Defaults to None.
```
measty, line 1067: This isnt clear. I'd suggest something like:
A geometry to query for annotations within (according to geometry_predicate). If None, all annotations in the store are considered. Defaults to None.

tiatoolbox/annotation/storage.py
John-P: @measty What does this comment mean? Is it a suggested rephrasing?

tiatoolbox/annotation/storage.py
John-P: @measty Is this better now? See the latest commit.

measty: Yes, looks fine now

measty:
Looks good! also tested it out with a real world example:
image
Glands with at least 1 neutrophil nearby

John-P:
> Looks good! also tested it out with a real world example: image Glands with at least 1 neutrophil nearby

Nice! This might be nice to put in the docs as an example.

tiatoolbox/annotation/storage.py
```diff
@@ -679,6 +686,26 @@ def query(
...
+            )
+
+        def bbox_centre_within(
+            annotation_geometry: Geometry, query_geometry: Geometry
+        ) -> bool:
+            """True if annotation bounds centre within query geometry bounds.
+
+            True if centroid of the annotation is within the query
+            geometry bounds.
+            """
```
shaneahmed, line 704:
```suggestion
            
            """
```

shaneahmed (approved):
Thanks @John-P. I have made an online commit to fix eol.
@measty Please let me know when you are happy with this PR. I will merge it then.

tiatoolbox/annotation/storage.py
```diff
@@ -1007,6 +1051,197 @@ def select_values(
...
+                geometry = Polygon.from_bounds(
+                    min_x - distance,
+                    min_y - distance,
+                    max_x + distance,
+                    max_y + distance,
+                )
+            elif from_mode == "boxpoint":
+                geometry_predicate = "bbox_centre_within"
+                geometry = Polygon.from_bounds(*ann.geometry.bounds).centroid
+                geometry = geometry.buffer(distance)
```
measty, line 1231: the above seems like it could be done more efficiently, as it will end up only using the bounds of the query geometry. The above essentially makes a box, gets its center, makes a bigger circle, then makes another box when the bbox_center_within is calculated. Instead, couldnt we:

```
#outside the for loop, make a box:  
query_box = box(-d/2, -d/2, d/2, d/2)  
#then instead of the above two lines  
bounds = ann.geometry.bounds  
geometry = translate(query_box, (bounds[2]-bounds[0])/2, (bounds[3]-bounds[1])/2)  
```
that would avoid the buffering of the point into a circle, which as it has to represent it as a polygon may be expensive.

measty (approved):
I've had another look through, and made one comment about the boxpoint-boxpoint implementation which may or may not help. But its purely to do with efficiency and not functionality, so am happy to approve this as is. If the change in the comment actually improves the benchmark, it could be incuded in this PR before its merged, or added in a separate one.

tiatoolbox/annotation/storage.py
John-P: Just checked and the predicate function (for DictionaryStore) is correct. It is checking if the centroid of the annotation geometry is inside the bounding box of the query geometry as passed to _query.

The mix-up here is that it is using the annotation geometry bounds and the query geometry centoid (which we needlessly make a centroid already here), not the other way around. This is due to how the spatial index functions.

John-P: The centroid used here is the centroid of the bounds and not the centroid of the geometry for rtree compatibility reasons, hence creating the bounds and then getting the centroid. The ann is unique for each loop iteration here, so cannot be moved outside the loop.

John-P: So here we get the bounds and then the centroid then buffer instead of simply centroid and then buffer because geometry.centroid != the centre point of the bounding box.

John-P: However, I do see that maybe this loop case should maybe the same as the above "bbox_intersects" except with the different geometry predict to be equivalent to SQLiteStore.

John-P: Thanks for raising this! I finally managed to work through it and fix the issue!

shaneahmed (approved):

measty (approved):
The updates look good, happy to approve

PR #543, opened 2023-02-17
:arrow_up: Upgrade `numpy` Dependency

shaneahmed:
- np.int, np.bool, np.float is depreciated in v1.20 in favour of np.int_, np.bool_, np.float_

Waiting on some dependencies to update:
- numba not yet compatible with numpy 1.24
  >   from numba.np.ufunc import _internal
  E   SystemError: initialization of _internal failed without raising an exception

   - https://github.com/numba/numba/issues/8464
   - https://github.com/numba/numba/pull/8691
   - https://github.com/numba/numba/issues/8841
   - https://github.com/numba/numba/milestone/63

- [x] Waiting for next numba release with numpy 1.24 support.


John-P (approved):

John-P:
Looks good but is it confirmed that this does not break map/numba compatibility?

PR #553, opened 2023-03-02
:arrow_up: DEV: Upgrade Scikit-image Dependency

shaneahmed:
- Upgrade Scikit-image Dependency to >=0.20
- selem is depericiated, use `footprint` instead


mostafajahanifar (approved):
All looks good to me. I tested this on Ubuntu and Windows and it installs just fine.

PR #556, opened 2023-03-08
✨ WIP: Add SlidingWindowInRegionsPatchExtractor

blaginin:
This PR adds `SlidingWindowInRegionsPatchExtractor`, which quickly samples regions of interest (ROI or regions) from an image into patches. 

Regions of interest are rectangles on an image that only need processing. For example, they are helpful if you train a pathology model on a bunch of images, and for each image, you only have specific annotated regions that can be used in learning. The concept of ROI is used when working with [Tumor-infiltrating lymphocytes](https://pubmed.ncbi.nlm.nih.gov/25214542/): for example, [the TIGER challenge](https://tiger.grand-challenge.org/Home/) contained train data as manually annotated regions of interest.

The task above can be solved with the existing `SlidingWindowPatchExtractor` with a manually created binary mask of whether a pixel is located in an ROI). However, this approach:
- requires **a lot** of memory: for each image, it creates and stores in RAM an array of all possible coordinate pairs that will be used for iteration;
- is slow because of the same reason;
- relies on the method which uses super unclear [magic numbers](https://github.com/TissueImageAnalytics/tiatoolbox/blob/0cbef86d62f87f5f4d8bb58351198c8f05831d9a/tiatoolbox/tools/patchextraction.py#L298) inside. 

Solving a simple ML task with `SlidingWindowPatchExtractor` required me an infinite number of time and memory. 
Thus, I wrote a new patch extractor for tasks that need fast data sampling and do not need mask support. 

https://user-images.githubusercontent.com/19199204/223590149-2f69957a-384d-4269-9c80-d7895f275273.mp4

---

The extractor does not store the coordinates inside but counts them on every request. That's why **it is much faster ⚡️:** [processing](https://gist.github.com/blaginin/01c518f2050370c083cffe9d21910d27) the same file takes 30 times less time and 450 times less memory than `SlidingWindowPatchExtractor`. 

To simplify data processing, I added support for coordinates extraction: if `return_coordinates` is set, the iterator will return a patch and its coordinates. For example, this can be useful to find the metadata objects in this patch. 

Because there exist many methods of sampling a region in batches, I added a few arguments to the class:
  - `within_region_bound` guarantees that all samples will be fully located inside the wsi;
- `within_wsi_bound` guarantees that all samples will be fully located inside the wsi. Same as in the original class;
- `min_region_covered` is the minimum number of pixels inside a region that all images have to have;
- `patch_size`, `resolution`, and `stride` from the original class are also supported. 

This image illustrates some of the arguments above:
wsi

### DIFF
--- a/tests/test_patch_extraction.py
+++ b/tests/test_patch_extraction.py
@@ -6,7 +6,10 @@ import numpy as np
 import pytest
 
 from tiatoolbox.tools import patchextraction
-from tiatoolbox.tools.patchextraction import PatchExtractor
+from tiatoolbox.tools.patchextraction import (
+    PatchExtractor,
+    SlidingWindowInRegionsPatchExtractor,
+)
 from tiatoolbox.utils import misc
 from tiatoolbox.utils.exceptions import FileNotSupported, MethodNotSupported
 from tiatoolbox.wsicore.wsireader import (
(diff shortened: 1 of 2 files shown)


blaginin:
Closing due to the new patch extraction method which is already fast enough

PR #557, opened 2023-03-08
:wrench: DEV: Filter Logger Output to `stdout` and `stderr`

shaneahmed:
- Filter logger output to `stdout` instead of `stderr`

### DIFF
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -56,8 +56,21 @@ def test_set_logger():
     assert len(logger.handlers) == 2
 
 
-def helper_logger_test(run_statement: str):
+def helper_logger_test(level: str):
     """Helper for logger tests."""
+    if level.lower() in ["debug", "info"]:
+        output = "out"
+        order = (0, 1)
+    else:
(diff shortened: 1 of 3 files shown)


John-P (approved):
Looks good. In future perhaps a custom steam handler class would be good to have errors on stderr but all other output on stdout.

shaneahmed:
> > Looks good. In future perhaps a custom steam handler class would be good to have errors on stderr but all other output on stdout. This would allow filtering error vs other output properly on Linux/UNIX/POSIX systems.
> 
> Great idea! I think I have implemented it here [a0db2ca](https://github.com/TissueImageAnalytics/tiatoolbox/commit/a0db2ca7797bf5cbbb479b981dc091624ddc1ccb) but struggling with the tests.

@John-P Finally found a way to test these here https://github.com/TissueImageAnalytics/tiatoolbox/pull/557/commits/9426e2cb2103e98ba5efd43e69f34b54a5c35537

John-P (approved):

PR #564, opened 2023-03-10
fix level when using dicom read_region

measty:
This PR fixes the dicom slide thunbnail issue (maybe). the issue seems to arise from a difference in how the wsidicom reader treats levels, in that it labels them by their relative magnification.

so the levels in the wsidicom reader are [0, 2, 4]

whereas in our WSIReader level is just an index [0, 1, 2], and the level_downsamples are separate.

My knowledge of DICOM is nonexistant, so it may be that the fix is wrong - hopefully john can have a look at it. It fixed the thumbnail issue for me though.

### DIFF
--- a/tiatoolbox/wsicore/wsireader.py
+++ b/tiatoolbox/wsicore/wsireader.py
@@ -4072,7 +4072,8 @@ class DICOMWSIReader(WSIReader):
         _, constrained_read_size = utils.transforms.bounds2locsize(
             constrained_read_bounds
         )
-        im_region = wsi.read_region(location, read_level, constrained_read_size)
+        dicom_level = wsi.levels[read_level].level
+        im_region = wsi.read_region(location, dicom_level, constrained_read_size)
         im_region = np.array(im_region)
 
         # Apply padding outside the slide area
@@ -4246,8 +4247,9 @@ class DICOMWSIReader(WSIReader):
             level_location, size_at_read_level, level_size
         )
         _, read_size = utils.transforms.bounds2locsize(read_bounds)
+        dicom_level = wsi.levels[read_level].level
         im_region = wsi.read_region(
-            location=location_at_baseline, level=read_level, size=read_size
+            location=location_at_baseline, level=dicom_level, size=read_size
         )
         im_region = np.array(im_region)
 


measty:
I've tested this on another example that john provided (a non-folder dicom), and it worked as expected, so i'm fairly confident this is ok, now its been tested on more than one example.

John-P:
I can't find where in the wsidicom source or documentation this behaviour is specified. Can you provide any refernce?

measty:
> I can't find where in the wsidicom source or documentation this behaviour is specified. Can you provide any reference? It's not really a DICOM thing but a wsidicom thing IMO.

Yes, its a wsidicom thing. it's not specifically documented, but you can see it in the source, for example:

```
    def calculate_scale(self, level_to: int) -> int:
        """Return scaling factor to given level.
        Parameters
        ----------
        level_to -- index of level to scale to
        Returns
        ----------
        int
            Scaling factor between this level and given level
        """
        return int(2 ** (level_to - self.level))

    def _assign_level(self, base_pixel_spacing: SizeMm) -> int:
        """Return (2^level scale factor) based on pixel spacing.
        Will round to closest integer. Raises NotImplementedError if level is
        to far from integer.
        Parameters
        ----------
        base_pixel_spacing: SizeMm
            The pixel spacing of the base lavel
        Returns
        ----------
        int
            The pyramid order of the level
        """
        float_level = math.log2(
            self.pixel_spacing.width/base_pixel_spacing.width
        )
        level = int(round(float_level))
        TOLERANCE = 1e-2
        if not math.isclose(float_level, level, rel_tol=TOLERANCE):
            raise NotImplementedError("Levels needs to be integer.")
        return level
```

from which you can see that the level is defined as the power of 2 between the pixel size at that level compared to base resolution, as opposed to being an index into a list of levels with the scale information stored independently from that as in tiatoolbox WSIReader. 

So what wsidicom is expecting as the level in read_region is log2(level_downsamples[read_level]) , not read_level. (or we can just get the level directly from the wsidicom level instance, as in the PR here)

John-P:
@measty thanks for finding that! Just wanted to verify the behaviour before we change it. Looks good.

shaneahmed (approved):
Thanks @measty

PR #568, opened 2023-03-17
🩹 BUG: Fix Parsing Missing Omero Version NGFF Metadata

John-P:
# To-Do

- [x] Handle missing version keys and null values.
- [x] Add test for missing value
- [x] Add test for non-numeric values e.g. "0.5-dev" [see current spec doc](https://ngff.openmicroscopy.org/latest/#omero-md)
- [x] Warn if version < or > 0.4
- [x] Warn for inconsistent versions

### DIFF
--- a/tiatoolbox/wsicore/wsireader.py
+++ b/tiatoolbox/wsicore/wsireader.py
@@ -97,7 +97,7 @@ def is_zarr(path: pathlib.Path) -> bool:
         return False
 
 
-def is_ngff(path: pathlib.Path, min_version: Tuple[int, ...] = (0, 4)) -> bool:
+def is_ngff(path: pathlib.Path, min_version: Tuple[str, ...] = ("0", "4")) -> bool:
     """Check if the input is a NGFF file.
 
     Args:
--- a/tiatoolbox/wsicore/wsireader.py
+++ b/tiatoolbox/wsicore/wsireader.py
@@ -139,10 +139,11 @@ def is_ngff(path: pathlib.Path, min_version: Tuple[int, ...] = (0, 4)) -> bool:
     omero_version = omero.get("verion")
     if omero_version:
         omero_version = tuple(omero_version.split("."))
+        if omero_version < min_version:
+            return False
     if any(version < min_version for version in multiscales_versions):
         return False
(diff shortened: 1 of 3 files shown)


tiatoolbox/wsicore/wsireader.py
```diff
@@ -136,11 +136,14 @@ def is_ngff(path: pathlib.Path, min_version: Tuple[int, ...] = (0, 4)) -> bool:
         tuple(int(part) for part in scale.get("version", "").split("."))
         for scale in multiscales
     )
-    omero_version = tuple(int(part) for part in omero.get("version", "").split("."))
+    omero_version = omero.get("verion")
```
shaneahmed, line 139:
```suggestion
    omero_version = omero.get("version")
```

tiatoolbox/wsicore/wsireader.py
```diff
@@ -136,11 +136,14 @@ def is_ngff(path: pathlib.Path, min_version: Tuple[int, ...] = (0, 4)) -> bool:
...
+    if omero_version:
```
shaneahmed, line 140: shouldn't it log a warning if this is empty?

John-P: It's perfectly valid to be empty according to the spec

tiatoolbox/wsicore/wsireader.py
```diff
@@ -36,6 +37,8 @@
 Bounds = Tuple[Number, Number, Number, Number]
 IntBounds = Tuple[int, int, int, int]
 Resolution = Union[Number, Tuple[Number, Number], np.ndarray]
+MIN_NGFF_VERSION = Version("0.4")
+MAX_NGFF_VERSION = Version("0.4")
```
shaneahmed, line 41: is this not quite strict?

John-P: Yes, intentionally. It's not been tested with anything other than 0.4 so we cannot say we suppport it

John-P: Besides, support for new versions is quite far beyond the scope of this PR.

PR #569, opened 2023-03-17
BUF: Fix docker files for ubuntu OSs

mostafajahanifar:
This docker fixes the bug with Ubuntu OS docker images which are because of `apt upgrade incompatibility.

### DIFF
--- a/docker/3.9/Ubuntu/Dockerfile
+++ b/docker/3.9/Ubuntu/Dockerfile
@@ -4,16 +4,14 @@ FROM ubuntu:22.04 AS builder-image
 ENV DEBIAN_FRONTEND=noninteractive
 
 # Install python3.9
-RUN : \
-    && apt-get update && apt upgrade \
-    && apt install software-properties-common -y \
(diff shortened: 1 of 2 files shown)


docker/3.9/Ubuntu/Dockerfile
```diff
@@ -31,4 +29,4 @@ RUN pip install --no-cache-dir tiatoolbox
 ENV VIRTUAL_ENV=/opt/venv
 ENV PATH="/opt/venv/bin:$PATH"
 
-CMD ["python"]
+# CMD ["python"]
```
shaneahmed, line 32: Why this link is commented?

mostafajahanifar: This is commented out to set the entry point to bash script. Otherwise, the docker container can only run Python commands. 
This change will make the functionality of the docker image more versatile.

shaneahmed (approved):
Thanks @mostafajahanifar

PR #571, opened 2023-03-22
⚡ Faster Filtering in Patch Dataset

measty:
* API Change for `PatchExtractor.filter_coordinates`. Instead of `coordinate_resolution`, `coordinate_units`, & `mask_resolution`, it takes `slide_shape` and a callable function `func` to validate the coordinates as input. New implementation is about ~200 times fast.

I've been using WSIPatchDataset for something recently, and noticed that it uses filter_coordinates method of PatchExtractor, instead of the faster filter_coordinates_fast. I can't see any reason why it shouldn't use the faster version - I am guessing that its just a case of, when the faster version was added, WSIPatchDataset was never updated to use it.

So, this PR updates it to use the faster method, and also includes a few other small improvements like allowing the min_mask_ratio argument to be used in the WSIPatchDataset in the same way it is available for patch extractor, so that behaviour between PatchExtractor and WSIPatchDataset is consistent.

Changes in this PR:
1. WSIPatchDataset now uses filter_coordinates_fast, consistent with PatchExtractor.
2. filter_coordinates_fast now uses the relative scale between the wsi dimensions at the requested patch resolution, and the mask dimensions, as the scale factor. This is units-agnostic and gives consistent results regardless of mode. Also doesn't make an even lower-res thumbnail of an already low-res mask as it did previously, which was unnecessary and could have had some wierd effects with masks which were already very low-res.
3. the documented post_proc argument now exists and will be used if given. Changed the documentation to reflect its functionality (ie its just called as is on the patch before it is returned)
4. Allows min_mask_ratio to be used in WSIPatchDataset, consistent with usage in PatchExtractor.

Rough performance comparison:
~0.6 sec using filter_coordinates_fast
~4 min using filter_coordinates
consistent with what was reported in original PR which added filter_coordinates_fast to PatchExtractor.

The PR also replicates the ability to provide a custom 'filter func' from filter_coordinates into filter_coordinates_fast, so in theory we can now get rid of filter_coordinates as it does nothing filter_coordinates_fast doesn't do.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,6 @@
+[tool.pytest.ini_options]
+  collect_ignore = ["setup.py", "benchmark/"]
+
 [tool.black]
   exclude = '''
   /(
@@ -34,3 +37,17 @@
   ]
   ignore_errors = true
   omit = ['tests/*', 'tiatoolbox/__main__.py', '*/utils/env_detection.py']
+
+[tool.isort]
+  profile = "black"
+  multi_line_output = 3
+  line_length = 88
+  filter_files = "True"
+  skip = [".gitignore", "benchmarks/*"]
+
+[build-system]
+  requires = ["setuptools"]
+  build-backend = "setuptools.build_meta"
+
+[tool.distutils.bdist_wheel]
+  universal = true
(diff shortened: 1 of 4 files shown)


blaginin:
It feels like a good idea to mark `filter_coordinates` as deprecated to remove it in the future versions

shaneahmed (approved):
Thanks @measty

PR #573, opened 2023-03-23
🔧 Use `pyproject.toml` for `pytest` Configuration

shaneahmed:
- Use `pyproject.toml` for `pytest` Configuration


John-P (approved):
We should probably add some options here, such as for coverage. Most of the time when running PyTest users will want coverage reported in the terminal (and XML/HTML). I suggest adding `--durations=10`, `--cov=tiatoolbox`, `--cov-report=term` and maybe also `--cov-report=xml` and `--cov-report=html`.

PR #577, opened 2023-03-31
✨ Return True for `is_ngff` With Warning Above Max NGFF Version

John-P:


shaneahmed:
I think you need to update tests.

PR #584, opened 2023-04-05
🚨 Use `logger` Instead Of `warnings` for `wsireader.py`

shaneahmed:
- Use `logger` Instead Of `warnings` for `wsireader.py`


measty (approved):
This looks fine to me

PR #587, opened 2023-04-10
🚨 Use `logger` Instead Of `warnings` for `wsi_registration.py`

shaneahmed:
- Use `logger` Instead Of `warnings` for `wsi_registration.py`
- Refactor duplicate code fragments

### DIFF
--- a/tiatoolbox/tools/registration/wsi_registration.py
+++ b/tiatoolbox/tools/registration/wsi_registration.py
@@ -86,6 +86,34 @@ def compute_center_of_mass(mask: np.ndarray) -> tuple:
     return (x_coord_center, y_coord_center)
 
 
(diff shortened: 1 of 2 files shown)


tiatoolbox/tools/registration/wsi_registration.py
```diff
@@ -86,6 +86,32 @@ def compute_center_of_mass(mask: np.ndarray) -> tuple:
     return (x_coord_center, y_coord_center)
 
 
+def apply_affine_transformation(fixed_img, moving_img, transform_initializer):
+    """Apply affine transformation using opencv.
```
shaneahmed, line 90:
```suggestion
    """Apply affine transformation using OpenCV.
```

PR #588, opened 2023-04-11
🚨 Use `logger` Instead Of `warnings` for `models` Package

shaneahmed:
- Use `logger` Instead Of `warnings` for `models` Package


tests/models/test_arch_nuclick.py
```diff
@@ -15,7 +14,7 @@
 # Test pretrained Model =============================
 
 
-def test_functional_nuclcik(remote_sample, tmp_path):
+def test_functional_nuclcik(remote_sample, tmp_path, caplog):
```
adamshephard, line 17: nuclick not nuclcik

PR #597, opened 2023-05-02
🐛 Fix `MultiTaskSegmentor` Example

blaginin:
- Fix `test_validate_docstring_examples`
- Revert to `test_validate_docstring_examples` execution in CI, as now it passes


adamshephard (changes requested):
It's probably more consistent to use '...' on the following line rather than using '\'.

adamshephard (approved):

PR #598, opened 2023-05-02
:recycle: Refactor `Resolution` and `Units`

blaginin:
- Removed multiple `Resolution` definitions
- Introduced `Units`
- Updated docstrings and type hints


John-P (approved):
Great suggestion. Happy with this. 👍

shaneahmed:
@blaginin are you still working on this PR as this is still a draft. You need to resolve conflicts as well before I merge.

blaginin:
@shaneahmed, you can review and merge. I got three points to emphasize:

1. In some cases, `units: str` → `units: Units` extended the scope of allowed options for this parameter. For example, PatchExtractor docstring said that _only_ `'mpp', 'power' or 'level'` are supported, but `Units` also have a `baseline` option. In cases like this, I checked that methods also support extended arguments. 

2. The only case I discovered where the point above is not true is PatchPredictor, where `units='level'` are not supported. However, then the `predict` method will crash with ```ValueError: Unknown units `level`. Units should be one of 'baseline', 'mpp' or 'power'.``` (which is good).

3. Two years ago @mostafajahanifar [added](https://github.com/TissueImageAnalytics/tiatoolbox/pull/105) a note to PatchExtractor that `Only pyramid / resolution levels (level) embedded in the whole slide image are supported`. Since then we've started reading patches in the [target resolution](https://github.com/TissueImageAnalytics/tiatoolbox/blob/80fb13c4aeda89db4de073d9d1e80de4da0ba8c4/tiatoolbox/tools/patchextraction.py#L190), so I removed this restriction.

tiatoolbox/wsicore/wsimeta.py
```diff
@@ -8,13 +8,14 @@
 """
 from numbers import Number
 from pathlib import Path
-from typing import List, Mapping, Optional, Sequence, Tuple, Union
+from typing import List, Literal, Mapping, Optional, Sequence, Tuple, Union
```
John-P, line 11: Glad we can use Literal now that we are 3.8+ 🚀

John-P (approved):
Looks great to me!

Issue #602, opened 2023-05-08
Issues with reproducing the result in examples/inference-pipelines/slide-graph.ipynb 


mohsuliman:
- TIA Toolbox version: 1.3.3
- Python version: 3.9.16
- Operating System: Ubuntu 18.04.6

### Description

I am trying to reproduce the results given at examples/inference-pipelines/slide-graph.ipynb using FEATURE_MODE = "composition". I have left everything in the code unchanged and set NUM_NODE_FEATURES =4 to match the given graph.

I have encountered the following two issues: 

1- Running the code unchanged gives me the following error in the code line that has "holder[uids] = freqs"

```
IndexError: arrays used as indices must be of integer (or boolean) type
```

Forcing the type of "uids" to be int seems to solve the issue, but I wonder why I have this error in the first place?

2- The second issue is that I am not getting a graph that matches the already provided graph. The network generates nodes with all zero features in the regions ruled out by the mask (i.e., where we shouldn't have any nodes) and assign edges to them as well. Hence, my final image graph has nodes (with edges) everywhere in the image. Also, as opposed to what is suggested in the notebook, the code's run time takes more than 7 hours. Am I missing something fundamental here?

shaneahmed:
Have you tried running the notebook on Colab? I just ran it on Colab and it runs fine for me without errors.

mohsuliman:
@shaneahmed  thanks for the reply. 

I am not sure if i can run it in Colab. Running the cell (In [12]) takes more than 4 hours and then Colab disconnects (as I don't have Colab Pro). Again running it in my local GPU it takes more than 7 hours. 

how long does it take to run the cell In [12]. It

mohsuliman:
@shaneahmed 

The only change that I have made in Colab is that i commented the part:
```
 # ! for using pre-generated graph
NUM_NODE_FEATURES = 4
FEATURE_MODE = "composition"
USE_PRE_GENERATED = True
```

and uncommented the part of:
```
# ! for trying out the function
NUM_NODE_FEATURES = 5
FEATURE_MODE = "composition"
USE_PRE_GENERATED = False
```

mohsuliman:
@shaneahmed  Yes, I have done all these steps and the issue is still there. Just tried it again and it went over 4 hours before Colab run stopped. What else could be the problem :/ 

I am not sure how is it taking 5 min on your Colab. It is mentioned here that "composition" may take over an hour (which is also not the case for me)
[...]

PR #608, opened 2023-05-16
:bug: Fix Errors in the `slidegraph` Example Notebook

measty:
Fixes some issues with the slidegraph notebooks.

1. Updates due to changes in STRtree in recent shapely versions
2. In the 'cell-composition' mode, add the 'filter_coordinates' step so that the mask is considered when generating graph nodes. Also made a small tweak to mask filter so that mask doesnt have to be single channel.
3. Fix the resolution of the plots being wrong when not using pre-generated model
4. Fixes for a couple of issues related to datatypes, maybe they crept in at some point due to numpy or torch version changes.

I have also added a note to explain the last few cells of the inference notebook are for composition features only, as there are only pretrained model weights for that mode.


shaneahmed:
Thanks @measty. Have you tested these notebooks to answer #602 ?

shaneahmed (approved):

PR #612, opened 2023-05-22
🐛 Fix Errors with `glymur` Update to `0.12.6`

shaneahmed:
- Fix Errors with `glymur` update


John-P (approved):
Looks like this is all sorted now 👍

PR #621, opened 2023-06-15
:recycle: Replace deprecated input `pretrained` with `weights`

shaneahmed:
- Replace deprecated input `pretrained` with `weights` according to the new API.


measty (approved):
Looks good

PR #622, opened 2023-06-15
:wastebasket: Remove `shapely.speedups`

shaneahmed:
- Remove `shapely.speedups`.
- [The shapely.speedups module (the enable and disable functions) is deprecated and will be removed in the future. The module no longer has any affect in Shapely >=2.0.](https://shapely.readthedocs.io/en/stable/release/2.x.html#shapely-2-0-api-changes-deprecated-in-1-8)


measty (approved):
looks fine to me

PR #626, opened 2023-06-30
:bug: Fix `dictionary changed size` Error

shaneahmed:
- Fix `dictionary changed size` error in 08-nucleus-instance-segmentation and 09-multi-task-segmentation
- Changing num_loader_workers=0 and num_postproc_workers=0 fixes the error on Colab.
- [09-multi-task-segmentation](https://github.com/TissueImageAnalytics/tiatoolbox/blob/develop/examples/09-multi-task-segmentation.ipynb) fails to load tmp\sample_tile_results\0.raw.0.npy still fails which is not saved.


shaneahmed:
[09-multi-task-segmentation](https://github.com/TissueImageAnalytics/tiatoolbox/blob/develop/examples/09-multi-task-segmentation.ipynb) fails to load tmp\sample_tile_results\0.raw.0.npy still fails which is not saved.

adamshephard (approved):

John-P:
I noticed that you have set the worker subprocesses to 0 in the notebooks. However, is this really addressing the root cause of the issue? I believe we should be able to use more workers in the notebooks. Is this just a temporary solution?

shaneahmed:
> I noticed that you have set the worker subprocesses to 0 in the notebooks. However, is this really addressing the root cause of the issue? I believe we should be able to use more workers in the notebooks. Is this just a temporary solution?

I have tested this locally with multiple Python versions on Windows and Ubuntu with num_workers=4, 16 and it works fine. Something failing with multiprocessing on Colab.

PR #629, opened 2023-07-03
:technologist: Reduce Test runtime for JP2

shaneahmed:
- Reduce Test runtime for JP2 `test_fuzz_read_region_baseline_size`
- This test takes longest time to run on CI. Number of iterations is quite high which can be reduced to improve tests run time.


John-P (approved):

John-P:
We could also get a smaller test JP2 image. That way we can use more iterations for a robust test but it should be faster.

shaneahmed:
> We could also get a smaller test JP2 image. That way we can use more iterations for a robust test but it should be faster.

This was the smallest we could get but I will try.

PR #638, opened 2023-07-07
:technologist: Use `poetry` for Version Updates

shaneahmed:
- Use [`poetry-bumpversion`](https://pypi.org/project/poetry-bumpversion/) for version number updates instead of bump2version as bump2version is no longer being maintained https://github.com/c4urself/bump2version/issues/42#issuecomment-1582436696.


John-P (approved):
Great!

Issue #603, opened 2023-05-08
🐌  Slow tests


blaginin:
- TIA Toolbox version: 1.4.0
- Python version:  3.10.8 
- Operating System: Mac OS 13.3.1 (arm)

### Description

Tests in the toolbox are extremely slow. Even with `CI=True`, all the checks take almost half an hour. Non-CI mode takes even more time. This makes an interactive development approach, with running tests after small changes, impossible. This results in writing [bad](https://github.com/blaginin/tiatoolbox/commit/267b91af795ccc548c8a19a9026004e97510bbfa#diff-5c6dddc0c85981eb75bdb49d400e82c9271918e138990baba53ac34cbcf2aaacR29) test-related code.

The main reasons why tests are slow are static files. Each test run is related to downloading and then deleting gigabytes of models' weights and test images. However, almost always, those files remain unchanged after the tests' execution. 

Not only is this slow, but it also triggers a lot of false positive results. For example, that's the typical result of my local run: 

<img width="1439" alt="image" src="https://user-images.githubusercontent.com/19199204/236876594-908c7835-704f-4ad6-938a-f761796b49dd.png">

# Todo

- [x] Fixtures caching
- [x] Models caching
- [x] Cache in CI
- [x] Parallel tests
- [ ] Safe downloads
- [ ] Cache rechecks (HTTP headers / ETag)

blaginin:
Of course, we want to be sure of the tests' reproducibility and don't want our code to fail because some code before has modified our assets. But maybe we can consider more user-friendly approaches? I have those options in mind: 
- Compare a file's local last change time with the remote one and download only if it's greater. This is done by `If-Range` / `Last-Modified` headers, which are already supported by `tiatoolbox.dcs.warwick.ac.uk` server. 
- Store files checksum on the server side and download a file only if its checksum is different
- Copy files from the local cache folder into the tests' sandbox each time tests are run

blaginin:
@John-P @shaneahmed, maybe you have other ideas?

shaneahmed:
@blaginin As John mentioned downloads should be for a session. Please let us know if this is not the case. Feel free to create a PR to fix this.

PR #625, opened 2023-06-22
✨ Use `ruff` for Linting

shaneahmed:
- Add checks for 
    - pydocstyle
    - flake8-future-annotations
    - flake8-errmsg
    - pygrep-hooks
    - flake8-use-pathlib
    - Ruff-specific rules
    - flake8-self
    - flake8-type-checking
    - tryceratops

PR #641, opened 2023-07-17
✅ Reuse models and datasets in tests

blaginin:
This PR addresses part of issue #603 by improving model weights and datasets in tests. Previously, weights for each model were downloaded to `f"{tmp_path}/weights.pth"` and replaced each other. The only available dataset, Kather, was downloaded into `TIATOOLBOX_HOME`, but the entire home directory was deleted during test execution. This approach has several drawbacks:

- Tests are extremely slow. On every test run, it downloads ~8 GB of weight files which never change.
- Tests cannot be run in parallel as they depend on the same `f"{tmp_path}/weights.pth"` file.
- Tests heavily load TIA's infrastructure. For each test run, the server has to transfer almost 10 GB of static assets. This might work while the number of contributors is small, but what happens when it increases?
- Tests were inconsistent with their models caching. Why, for example, PatchPredictor cached their pretrained_models, but UNetModel did not?

Now, all weights are downloaded to TIATOOLBOX_HOME and cached between runs. Storing more data in memory (+8 GB) makes tests much faster, especially if you're not in Coventry or have slow internet. In rare cases where you specifically need to test behaviour without cache, you can set TIATOOLBOX_HOME to a temporary directory.


tiatoolbox/models/architecture/__init__.py
```diff
@@ -29,9 +31,22 @@ def fetch_pretrained_weights(model_name: str, save_path: str, overwrite: bool =
...
+
     """
+    if model_name not in PRETRAINED_INFO:
+        raise ValueError(f"Pretrained model `{model_name}` does not exist")
+
     info = PRETRAINED_INFO[model_name]
+
+    if save_path is None:
+        file_name = info["url"].split("/")[-1]
+        save_path = os.path.join(rcParam["TIATOOLBOX_HOME"], "models/", file_name)
```
shaneahmed, line 46: rcParam["TIATOOLBOX_HOME"] is now a Path in #625. You can use rcParam["TIATOOLBOX_HOME"] / "models" / file_name

shaneahmed (approved):
Thanks @blaginin

PR #640, opened 2023-07-16
:white_check_mark: Upgrade remote sample fixtures

blaginin:
This PR addresses part of issue #603 and improves fixtures in tests:

- Some tests were using `_fetch_remote_sample` even though tests have a special `remote_sample` wrapper.
- `remote_sample` previously created a new folder for each run (data-0, data-1, data-2, and so on). Now it uses just one folder for all remote fixtures during a test session.
- To prevent duplicated downloads, I added download locks to the fixtures.
- I combined `_fetch_remote_sample` and `download_data` since they had the same logic.

PR #644, opened 2023-07-19
💚 Add CI Cache

blaginin:


shaneahmed:
> Download and computation times vary a lot, but here's a quick snippet of the two runs:
> 
> [No cache + Make one](https://github.com/blaginin/tiatoolbox/actions/runs/5626117463/attempts/2)	[Use made cache 💨 💨 💨 ](https://github.com/blaginin/tiatoolbox/actions/runs/5626117463/attempts/3)
> image	image
> Skipped 64 downloads (repeated assets during the run) thanks to #640	Skipped 79 downloads (repeated assets + model weights, ~6GB)
> 41 downloads (assets + models weights)	26 downloads (assets loading the first time, ~3GB)
> I don't cache:
> 
> * First-time loading assets (not thread safe yet, but maybe finish in [💚 Add cache for the static assets #674](https://github.com/TissueImageAnalytics/tiatoolbox/pull/674)
> * PyTorch Models (downloads are extremely fast already, no need to cache)

I agree there is a difference, although it is not easy to measure due to inconsistencies in the way GitHub actions work due to load. I will merge it considering that there is an improvement.

shaneahmed (approved):

PR #645, opened 2023-07-19
Add check for Phase correlation transform

ruqayya:
Likewise other intermediate transform, Phase cross-correlation transform should only be applied if it improves the DICE overlap.

### DIFF
--- a/tiatoolbox/tools/registration/wsi_registration.py
+++ b/tiatoolbox/tools/registration/wsi_registration.py
@@ -1111,7 +1111,7 @@ class DFBRegister:
             fixed_tissue_mask, moving_tissue_mask, translation_offset
         )
 
-        # Use the estimated phase cross correlation transform 
+        # Use the estimated phase cross correlation transform
         # only if it improves DICE overlap
         after_dice = dice(fixed_tissue_mask, transform_tissue_mask)
         if after_dice < before_dice:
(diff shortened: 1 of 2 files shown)


tiatoolbox/tools/registration/wsi_registration.py
```diff
@@ -1107,6 +1107,14 @@ def register(
             fixed_tissue_img, moving_tissue_img
         )
         translation_offset = np.array([[1, 0, shift[1]], [0, 1, shift[0]], [0, 0, 1]])
+        transform_tissue_mask = apply_affine_transformation(
+            fixed_tissue_mask, moving_tissue_mask, translation_offset
+        )
+
+        # Use the estimated phase cross correlation transform only if it improves DICE overlap
```
shaneahmed, line 1114:
```suggestion
        # Use the estimated phase cross correlation transform 
        # only if it improves DICE overlap
```

shaneahmed (commented):
I think as there is an if statement you may need to add a test. 
You can create a new function that would help to add a quick test.

ruqayya:
> I think as there is an if statement you may need to add a test. You can create a new function that would help to add a quick test.

'If statement' is already covered by the existing test functions such as test_register_output_with_initializer. The phase cross correlation transform for the images that we are using for testing results in low dice value as compared to previous dice value and hence "if statement" is executed on calling register function. Do we still need to have a separate function?

shaneahmed (approved):
Thanks @ruqayya

PR #651, opened 2023-07-22
🐛 Parse JP2 Boxes By ID Not Position

John-P:
Boxes can be in any order. Previously the order was being relied on.

### DIFF
--- a/tests/test_wsireader.py
+++ b/tests/test_wsireader.py
@@ -1432,7 +1432,7 @@ def test_wsireader_open(
 def test_jp2_missing_cod(sample_jp2, caplog):
     """Test for warning if JP2 is missing COD segment."""
     wsi = wsireader.OmnyxJP2WSIReader(sample_jp2)
-    wsi.glymur_wsi.codestream.segment = []
+    wsi.glymur_jp2.codestream.segment = []
     _ = wsi.info
     assert "missing COD" in caplog.text
 
--- a/tiatoolbox/wsicore/wsireader.py
+++ b/tiatoolbox/wsicore/wsireader.py
@@ -10,7 +10,7 @@ import pathlib
 import re
 from datetime import datetime
 from numbers import Number
-from typing import Iterable, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Iterable, List, Optional, Tuple, Union
 
 import numpy as np
 import openslide
@@ -30,6 +30,9 @@ from tiatoolbox.utils.magic import is_sqlite3
 from tiatoolbox.utils.visualization import AnnotationRenderer
 from tiatoolbox.wsicore.metadata.ngff import Multiscales
 from tiatoolbox.wsicore.wsimeta import WSIMeta
+
+if TYPE_CHECKING:
+    import glymur
 
 pixman_warning()
 
@@ -2088,7 +2091,7 @@ class OmnyxJP2WSIReader(WSIReader):
         import glymur
 
         glymur.set_option("lib.num_threads", os.cpu_count() or 1)
-        self.glymur_wsi = glymur.Jp2k(filename=str(self.input_path))
+        self.glymur_jp2 = glymur.Jp2k(filename=str(self.input_path))
 
     def read_rect(
         self,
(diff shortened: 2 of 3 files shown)


tests/test_wsireader.py
```diff
@@ -2319,6 +2325,89 @@ def test_ngff_inconsistent_multiscales_versions(tmp_path, caplog, remote_sample)
...
+    )
+    xml_path = tmp_path / "data.xml"
+    xml_path.write_text("<Jp2></Jp2>")
+    xmlbox = glymur.jp2box.XMLBox(filename=xml_path)
+    jp2.append(xmlbox)
+    _ = WSIReader.open(path).info
+
+
+def test_jp2_empty_xml_empty_description(tmp_path):
+    """Test WSIReader init for a JP2 with an empty Jp2 XML description tag."""
```
shaneahmed, line 2353: Is this JP2 or Jp2? It would be good to be consistent.

tests/test_wsireader.py
```diff
@@ -2319,6 +2325,89 @@ def test_ngff_inconsistent_multiscales_versions(tmp_path, caplog, remote_sample)
...
+    xmlbox = glymur.jp2box.XMLBox(filename=xml_path)
+    jp2.append(xmlbox)
+    _ = WSIReader.open(path).info
+
+
+def test_jp2_no_header(tmp_path, monkeypatch):
+    """Test WSIReader with an image crafted with a missing header.
+
+    Note: This should not be possible for a real file.
+    """
```
shaneahmed, line 2384:
```suggestion

    """
```

shaneahmed (commented):
@John-P Looks good. Need to be tested with `CMU-1-small.jp2`. We can merge it then.

tests/test_wsireader.py
John-P: It's JP2

tiatoolbox/wsicore/wsireader.py
```diff
@@ -2599,20 +2666,55 @@ def _info(self):
...
-        glymur_wsi = self.glymur_wsi
-        box = glymur_wsi.box
-        description = box[3].xml.find("description")
-        matches = re.search(r"(?<=AppMag = )\d\d", description.text)
-        objective_power = np.int_(matches[0])
-        image_header = box[2].box[0]
+        jp2 = self.glymur_jp2
+        boxes = self._get_jp2_boxes(jp2)
+        objective_power = None
+        vendor = None
```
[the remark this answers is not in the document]
John-P: I don't think that is a good idea. I believe returning None makes more semantic sense. E.g. what if the vendor field is "Not Available" then this is confusing. Returning None has semantic meaning that it was not present and fits in with standard Python bool checking etc.

tests/test_wsireader.py
shaneahmed: Should it be JP2 XML?

John-P: Yep

John-P:
@shaneahmed Can you make a new CVS for the patch extraction for the smaller JP2? This is the slowest test with JP2 now but cannot use the new image because the CVS is no longer valid.

shaneahmed:
> @shaneahmed Can you make a new CSV for the patch extraction for the smaller JP2? This is the slowest test with JP2 now but we cannot use the new image because the CSV is no longer valid.

OK I will add this in a separate PR.

John-P:
> OK I will add this in a separate PR.

I am confused. This PR will not pass until this is done. It just needs field upload rather than code changes. I can revert the change to the new file for the tests if you want it to be in a new PR but it is already done here. Just needs the new CSV and YAML files etc.

shaneahmed:
> > OK I will add this in a separate PR.
> 
> I am confused. This PR will not pass until this is done but doing just this part on the other PR will also fail without these changes. It just needs files upload rather than code changes. I can revert the change to the new file for the tests if you want it to be in a new PR but it is already done here. Just needs the new CSV and YAML files etc.

OK didn't see that as the tests were still running. I will upload a separate file after running tests locally.

John-P:
@shaneahmed Thank you for sorting out those files. Looks like something else is causing a problem now 😵

shaneahmed:
> @shaneahmed Thank you for sorting out those files. Looks like something else is causing a problem now 😵

I haven't uploaded the new files yet. May be it is failing as it is trying to access a location which is not available. I am testing this locally.

John-P:
Passing now 🎉

PR #678, opened 2023-08-01
:rotating_light: `ruff` tiatoolbox root

shaneahmed:
- Add `ruff` checks tiatoolbox root


tiatoolbox/__init__.py
```diff
@@ -12,6 +13,9 @@
 
 import yaml
 
+if TYPE_CHECKING:
```
shaneahmed, line 16:
```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

PR #721, opened 2023-10-10
✨  Add Support for `QuPath` Annotation Imports

measty:
This PR is intended to make importing objects from QuPath (and other applications) easier. I have updated it to follow the suggestion of John to do it via dependency injection.

If you export some objects in .geojson from some versions of QuPath, and they have measurements, all the measurements are in a 'measurements' property in the geojson. This means, when you import into annotation store, it will import fine but the properties dict will look something like:

`properties = {a couple other props, "measurements": [{'name': 'propA', 'value': valA}, {'name': 'propB', 'value': valB}, etc]}`

Which is awkward for many downstream things with the nested data structure, and doesn't really mesh with how annotations are intended to be represented in the store (usually a flat dict of properties unless there's a very good reason for nesting).

This PR adds an option to provide a transform to deal with any application-specific formatting, which for example can be used to unpack measurements when importing from QuPath to give instead in a properties dict with the measurement properties unpacked:

`properties = {a couple other props, 'propA': valA, 'propB': valB, etc}`

Importing from QuPath is probably one of the main use-cases for this, and an example is provided in the docstring for this case, but the additional flexibility means it can be used for other use-cases too.


shaneahmed (approved):
Thanks @measty

tiatoolbox/annotation/storage.py
```diff
@@ -1765,6 +1801,11 @@ def add_from_geojson(
                 at non-baseline resolution.
             origin (Tuple[float, float]):
                 The x and y coordinates to use as the origin for the annotations.
+            import_transform (Callable):
```
shaneahmed, line 1804:
```suggestion
            transform(Callable):
```

tests/test_annotation_stores.py
```diff
@@ -2823,3 +2823,65 @@ def test_query_min_area(
...
+                            {"name": "Area µm^2", "value": 27.739423751831055},
+                        ],
+                    },
+                },
+            ],
+        }
+        with (tmp_path / "test_annotations.geojson").open("w") as f:
+            json.dump(anns, f)
+
+        def unpack_qpath(ann: Annotation) -> Annotation:
```
shaneahmed, line 2868:
```suggestion
        def unpack_qupath(ann: Annotation) -> Annotation:
```

tiatoolbox/annotation/storage.py
shaneahmed: Fixed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/721/commits/dd16891f21c00639583e2fdd7817fb0b6c06621f

PR #723, opened 2023-10-16
:construction: Add type checking using `mypy`

Jiaqi-Lv:
Add type checking using `mypy`, a static type checker for Python.

This PR involves adding type hints and checking the correctness of existing type hints in the code base.
A GitHub action has been created to run `mypy` after push and pull requests.
Currently, only `tiatoolbox/__init__.py` has been checked to work without errors with `mypy`.

### DIFF
--- a/tiatoolbox/__init__.py
+++ b/tiatoolbox/__init__.py
@@ -9,7 +9,8 @@ from typing import TYPE_CHECKING, Dict
 if sys.version_info >= (3, 9):  # pragma: no cover
     import importlib.resources as importlib_resources
 else:  # pragma: no cover
-    import importlib_resources  # type: ignore[import-not-found] To support Python 3.8
+    # To support Python 3.8
+    import importlib_resources  # type: ignore[import-not-found]
 
 import yaml
 
--- a/tiatoolbox/__init__.py
+++ b/tiatoolbox/__init__.py
@@ -9,7 +9,7 @@ from typing import TYPE_CHECKING, Dict
 if sys.version_info >= (3, 9):  # pragma: no cover
     import importlib.resources as importlib_resources
 else:  # pragma: no cover
-    import importlib_resources  # To support Python 3.8
+    import importlib_resources  # type: ignore[import-not-found]  # To support Python 3.8
 
 import yaml
 
--- a/tiatoolbox/__init__.py
+++ b/tiatoolbox/__init__.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import importlib.util
 import sys
 from pathlib import Path
-from typing import TYPE_CHECKING, Dict
+from typing import TYPE_CHECKING
 
 if sys.version_info >= (3, 9):  # pragma: no cover
     import importlib.resources as importlib_resources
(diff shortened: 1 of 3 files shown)


tiatoolbox/__init__.py
```diff
@@ -88,8 +91,10 @@ def read_registry_files(path_to_registry: str | Path) -> dict | str:
 
 
     """
+    if isinstance(path_to_registry, Path):
```
shaneahmed, line 94:
```suggestion
```

You do not need an if statement here. "path_to_registry = str(path_to_registry)" should be enough.

tiatoolbox/__init__.py
```diff
@@ -101,14 +106,16 @@ def read_registry_files(path_to_registry: str | Path) -> dict | str:
...
     spec = importlib.util.spec_from_file_location(name, module_location)
+    if spec is None or spec.loader is None:
+        raise ModuleNotFoundError(name=name, path=str(module_location))
     loader = importlib.util.LazyLoader(spec.loader)
     spec.loader = loader
     module = importlib.util.module_from_spec(spec)
     sys.modules[name] = module
     loader.exec_module(module)
-    return module
+    return {name: module}
```
shaneahmed, line 118: What's the benefit of returning this as a dictionary?

tiatoolbox/__init__.py
Jiaqi-Lv: Yes `Path` here causes mypy to fail. It seems that you cannot use `/` on `Path` objects, `/` only works for `PurePath` objects.

tiatoolbox/__init__.py
Jiaqi-Lv: The return type for this function used to be `sys.modules`, but `sys.modules` is a variable so we can't use it as type. I looked up the type of `sys.modules`, it was `dict[str, ModuleType]` so I changed the return type to this, don't think there is any benefit of returning a dictionary.

Alternatively we change the return type of this function to `ModuleType` and keep everything else the same.

tiatoolbox/__init__.py
```diff
@@ -71,8 +73,9 @@ def filter(self: DuplicateFilter, record: LogRecord) -> bool:
 
 
 # runtime context parameters
+rcParam: dict[str, str | Path | dict]  # noqa: N816
 rcParam = {  # noqa: N816
-    "TIATOOLBOX_HOME": Path.home() / ".tiatoolbox",
+    "TIATOOLBOX_HOME": Path.home() / Path(".tiatoolbox"),
```
shaneahmed, line 78:
```suggestion
    "TIATOOLBOX_HOME": Path.home() / ".tiatoolbox",
```

tiatoolbox/__init__.py
```diff
@@ -88,6 +91,7 @@ def read_registry_files(path_to_registry: str | Path) -> dict | str:
 
 
     """
+    path_to_registry = str(path_to_registry)
```
shaneahmed, line 94:
```suggestion
    path_to_registry = str(path_to_registry)   # To pass tests with Python 3.8
```

shaneahmed (approved):

PR #731, opened 2023-10-31
:hammer: Add `typing.py` and `env_detection.py` to `mypy` Type Check

Jiaqi-Lv:
- Fix type error in `tiatoolbox/typing.py` and `tiatoolbox/utils/env_detection.py`

### DIFF
--- a/tiatoolbox/typing.py
+++ b/tiatoolbox/typing.py
@@ -1,28 +1,28 @@
 """Define Variable types for TIAToolbox."""
 from __future__ import annotations
 
-from typing import Callable, Dict, List, Literal, Sequence, SupportsFloat, Tuple
+from typing import Callable, Dict, List, Literal, Sequence, SupportsFloat, Tuple, Union
 
 import numpy as np
 from shapely.geometry import LineString, Point, Polygon  # type: ignore[import-untyped]
 
 # Proper type annotations for shapely is not yet avaliable
 
-JSON = Dict[str, "JSON"] | List["JSON"] | str | int | float | bool | None
+JSON = Union[Dict[str, "JSON"], List["JSON"], str, int, float, bool, None]
 NumPair = Tuple[SupportsFloat, SupportsFloat]
 IntPair = Tuple[int, int]
 
 # WSIReader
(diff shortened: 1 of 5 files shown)


tiatoolbox/utils/env_detection.py
```diff
@@ -365,7 +362,7 @@ def pixman_versions() -> tuple[list, str]:
             determined.
 
     """
-    versions = []
+    versions: list[Any] = []
```
shaneahmed, line 365: shouldn't this be list[int] ? as it says in docstring?

Jiaqi-Lv: Yeah it should be list[int]

tiatoolbox/typing.py
```diff
@@ -1,24 +1,26 @@
 """Define Variable types for TIAToolbox."""
-from typing import Callable, Literal, Sequence, SupportsFloat
+from typing import Callable, Dict, List, Literal, Sequence, SupportsFloat, Tuple, Union
```
shaneahmed, line 2:
```suggestion
from __future__ import annotations
from typing import Callable, Dict, List, Literal, Sequence, SupportsFloat, Tuple, Union
```

tiatoolbox/utils/env_detection.py
```diff
@@ -344,12 +340,12 @@ def check_pixman_using_macports(versions: list) -> tuple[list, str]:
         flags=re.MULTILINE,
     )
     if matches:
-        versions = [version_to_tuple(matches.group(1))]
+        versions = [version_to_tuple(match) for match in matches]
```
Jiaqi-Lv, line 343: `re.finall()` returns a list of strings, not a `match` object, so it doesn't have the  `group` attribute.

Jiaqi-Lv:
It seems like we **have** to use `Union` instead of `|` in type annotations, otherwise Python 3.8 and Python 3.9 won't pass the check, (even with `from __future__ import annotations`).

shaneahmed (approved):
Thanks @Jiaqi-Lv

Issue #735, opened 2023-11-06
example 05 is broken


measty:
- TIA Toolbox version: develop
- Python version: colab
- Operating System:

### Description

05 example notebook fails when run in colab

### What I Did

Opened example 05 in colab and ran it (after restarting runtime once toolbox etc is installed as per instructions)

### Reason

Some changes to overlay_prediction_mask() to make a previously non-required argument to now be a required keyword argument (return_ax) mean this notebook no longer works.

shaneahmed:
Does it work with the stable release? If not then we need a hotfix otherwise we can fix it on the develop branch.

measty:
it seems to work on release. if i replace the tiatoolbox install line in the notebook with:
`!pip install git+https://github.com/TissueImageAnalytics/tiatoolbox.git@v1.4.1 | tail -n 1`
it will run correctly

measty:
fixed in #736

PR #736, opened 2023-11-08
📝 Fix Example 05

measty:
Fixes issue #735


shaneahmed (approved):
Thanks @measty for fixing this.

PR #737, opened 2023-11-10
✨ Add WSI processing to `EngineABC`

AbishekRajVG:
Implements cache mode for processing using EngineABC. Cache_mode saves intermediate results to zarr file which can be converted to AnnotationStore. Full WSI pipeline needs to be implemented for Engines inheriting the base class.

The output of models should be a dictionary according to the new Engine design.


tests/engines/test_engine_abc.py
```diff
@@ -32,17 +34,48 @@ def __init__(
         """Test EngineABC init."""
         super().__init__(model=model, weights=weights, verbose=verbose)
 
-    def infer_wsi(self: EngineABC) -> NoReturn:
-        """Test infer_wsi."""
-        ...  # dummy function for tests.
-
-    def post_process_wsi(self: EngineABC) -> NoReturn:
+    def set_dataloader(
```
shaneahmed, line 37:
```suggestion
    def get_dataloader(
```

tiatoolbox/utils/misc.py
```diff
@@ -1217,19 +1217,19 @@ def dict_to_store(
...
     patch_coords = np.array(patch_output.get("coordinates", []))
     if not np.all(np.array(scale_factor) == 1):
         patch_coords = patch_coords * (np.tile(scale_factor, 2))  # to baseline mpp
     labels = patch_output.get("labels", [])
     # get classes to consider
-    if len(class_probs) == 0:
-        classes_predicted = np.unique(preds).tolist()
-    else:
-        classes_predicted = range(len(class_probs[0]))
+    classes_predicted = np.unique(preds).tolist()
```
AbishekRajVG, line 1226: Requesting @measty's review on this change:

While testing AnnotationStore for the WSI output, I used the existing `mini_wsi_svs = Path(sample_wsi_dict["wsi2_4k_4k_svs"])` image as input to the updated Engine ABC's `run()` method. 

The Post process step now has an extra parameter, `str: output_type` which can be "AnnotationStore", which calls the `utils.misc.dict_to_store()` method, which throws a random `Key Error` every time. Upon debugging, the `classes_predicted = np.unique(preds).tolist()` resolves it as in this given case, the classes_predicted is not necessarily all the numbers in 0 to 9. Please advise if we should remove that if check, or if this is expected behaviour. 

```============================================================================================== PASSES ===============================================================================
_______________________________________________________________________________ test_engine_run_wsi_annotation_store ________________________________________________________________________________
--------------------------------------------------------------------------------------- Captured stdout call ----------------------------------------------------------------------------------------
|2023-11-27|13:34:26.736| [INFO] When providing multiple whole slide images, the outputs will be saved, and the locations of outputs will be returned to the calling function.
[DEBUG] classes_predicted  [0, 1, 2, 3, 4, 6, 7]
[DEBUG] class_probs shape  (69, 9)
[DEBUG] preds shape  (69,)
[DEBUG] preds  [7, 3, 2, 3, 3, 3, 3, 4, 4, 3, 3, 3, 3, 7, 3, 3, 6, 4, 3, 4, 3, 3, 3, 4, 6, 0, 0, 0, 0, 3, 0, 0, 3, 3, 3, 0, 2, 3, 3, 3, 1, 3, 6, 2, 3, 3, 3, 3, 3, 0, 4, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 7, 3, 3, 3, 3, 3, 3, 2]
[DEBUG] labels  []
[Debug 2] Labels  None
[DEBUG] class_dict  {0: 0, 1: 1, 2: 2, 3: 3, 4: 4, 6: 6, 7: 7}
[DEBUG] class_dict shape  ()```

Comment (measty) [reply tiatoolbox/utils/misc.py:1226]: Hi, 
The classes_predicted cannot be changed that way, because when there are probabilities output, there will be predicted probabilities even for classes that don't appear in preds, and these will need to be saved.

The problem is rather in this part:
```
    if class_dict is None:
        # if no class dict create a default one
        class_dict = {i: i for i in np.unique(preds + labels).tolist()}
```
which needs to be modified to take into account the possibility that some classes for which a probability is output may not have an instance where that class 'wins' resulting in an instance classified as that class in preds.
Hopefully, the following modification would solve the issue:
```
    if class_dict is None:
        # if no class dict create a default one
        if len(class_probs) == 0:
              class_dict = {i: i for i in np.unique(preds + labels).tolist()}
        else:
              class_dict = {i: i for i in range(len(class_probs))}
```

Comment (AbishekRajVG) [reply tiatoolbox/utils/misc.py]: Hey Mark. Thank you, it is resolved now.

**Comment by shaneahmed** [comment]
Deepsource is failing for complexity. You can move this for loop in a private function to reduce complexity.

https://github.com/TissueImageAnalytics/tiatoolbox/blob/a5a44e604ca082c580df406ec5ab6d6bcdb962ac/tiatoolbox/utils/misc.py#L1244

File: tests/engines/test_engine_abc.py [hunk 3]
```diff
@@ -416,3 +431,237 @@ def test_patch_pred_zarr_store(tmp_path: pytest.TempPathFactory) -> NoReturn:
...
+        stride_shape=(256, 256),
+        input_resolutions=[{"resolution": 1.35, "units": "mpp"}],
+    )
+    eng.run(
+        [mini_wsi_svs],
+        patch_mode=False,
+        save_dir=f"{tmp_path}/dump",
+        ioconfig=ioconfig,
+    )
+    shutil.rmtree(tmp_path / "dump", ignore_errors=True)
```
shaneahmed, line 476: Is this testing the output before removing it?

tests/engines/test_engine_abc.py
```diff
@@ -416,3 +431,237 @@ def test_patch_pred_zarr_store(tmp_path: pytest.TempPathFactory) -> NoReturn:
...
+
+    eng.run(
+        [mini_wsi_svs],
+        patch_mode=False,
+        save_dir=f"{tmp_path}/dump",
+        **kwargs,
+    )
+    shutil.rmtree(tmp_path / "dump", ignore_errors=True)
```
shaneahmed, line 484: same as above.

tests/engines/test_engine_abc.py
```diff
@@ -416,3 +431,237 @@ def test_patch_pred_zarr_store(tmp_path: pytest.TempPathFactory) -> NoReturn:
...
+        [mini_wsi_svs],
+        units="baseline",
+        patch_mode=False,
+        save_dir=f"{tmp_path}/dump",
+    )
+    assert eng._ioconfig.input_resolutions[0]["units"] == "baseline"
+    shutil.rmtree(tmp_path / "dump", ignore_errors=True)
+
+
+def test_engine_run_wsi(
```
shaneahmed, line 526: Need to check if the output values are correct. For probabilities you can test they are in a particular range.

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -274,16 +269,41 @@ def _initialize_model_ioconfig(
...
+        masks: Path | None = None,
+        labels: list | None = None,
+        ioconfig: ModelIOConfigABC | None = None,
     ) -> torch.utils.data.DataLoader:
         """Pre-process an image patch."""
         if labels:
             # if a labels is provided, then return with the prediction
             self.return_labels = bool(labels)
 
+        if isinstance(images, Path):
```
shaneahmed, line 285: This should be checking the mode. 
WSI can also be provided as numpy array similar to patches. The difference is that the size of patches is equal to size required by the network architecture.

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -395,20 +415,253 @@ def setup_patch_dataset(
...
+                batch_data["image"],
+                device=self.device,
+            )
+            # We get the index of the class with the maximum probability
+            batch_output_predictions = self.model.postproc_func(
+                batch_output_probabilities,
+            )
+
+            # tolist might be very expensive
+            cum_output["probabilities"].extend(batch_output_probabilities.tolist())
```
shaneahmed, line 574: I think these values should be directly written to zarr otherwise it will create an overhead and complexity specially when we extend to more complex algorithms.

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -395,20 +415,253 @@ def setup_patch_dataset(
...
+        outputs = [cum_output]  # assign to a list
+
+        # pop unused items from cum_output
+        if len(cum_output["probabilities"]) == 0:
+            cum_output.pop("probabilities")
+        if not return_labels or len(cum_output["labels"]) == 0:
+            cum_output.pop("labels")
+
+        merged_prediction = None
+        if merge_predictions:
```
shaneahmed, line 598: I am not sure if we should keep it here, we can discuss this tomorrow.

tests/engines/test_engine_abc.py
```diff
@@ -416,3 +431,237 @@ def test_patch_pred_zarr_store(tmp_path: pytest.TempPathFactory) -> NoReturn:
...
+        masks=[mini_wsi_msk],
+        patch_mode=False,
+        output_type="AnnotationStore",
+        **kwargs,
+    )
+
+    for output_info in out.values():
+        assert Path(output_info).exists()
+        assert output_info.suffix == ".db"
+    shutil.rmtree(save_dir)
```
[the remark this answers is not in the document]
AbishekRajVG: This is updated now. I've written a helper function to extract the probabilities from the '.db' file and test if they're between 0,1.

tests/engines/test_engine_abc.py
AbishekRajVG: Updated. I've written a helper function to verify probabilities. For others, I've written checks to verify if the type is a Zarr array or to check if they're of the right dimension. Will cross-verify during the call.

tiatoolbox/models/engine/engine_abc.py
AbishekRajVG: [Update] This is partially improved. I've added another helper function to convert/persist the ndarrays as zarr arrays directly. 
[Pending] Instead of `.extend()` which eventually converts it to a list of zarr arrays which will again be inefficient. I'm going to have them as a zarr group. But this breaks the existing dictionary to Annotation store code.

AbishekRajVG: [Update] Fully improved. I've added the `wsi_batch_output_to_zarr_group()` function in `misc.py` the saves the intermediate batch outputs of TIAToolbox engines directly to a Zarr file
Full diff: [[commit 9344c0639b](https://github.com/TissueImageAnalytics/tiatoolbox/pull/737/commits/9344c0639b30efc40654d2011b677a8ed1d4067d)]

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -396,19 +423,126 @@ def setup_patch_dataset(
...
+        dataloader: torch.utils.data.DataLoader,
+        img_label: str,
+        highest_input_resolution: list[dict],
+        save_dir: Path,
+        **kwargs: dict,
+    ) -> list:
         """Model inference on a WSI."""
-        raise NotImplementedError
+        # return coordinates of patches processed within a tile / whole-slide image
+        return_coordinates = True
```
shaneahmed, line 436: What's the purpose of this variable if it is always True?

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -612,25 +819,78 @@ def run(
...
-        return {"save_dir": save_dir}
+        self._ioconfig = self._update_ioconfig(
+            ioconfig,
+            self.patch_input_shape,
+            self.stride_shape,
+            self.resolution,
+            self.units,
+        )
+
+        fx_list = self._ioconfig.scale_to_highest(
```
shaneahmed, line 851:
```suggestion
        scale_factors = self._ioconfig.scale_to_highest(
```

measty:
As a general comment, I feel like the EngineABC does 'too much'. Some of the implemented methods are specific to a PatchPrediction engine, and would not be relevant for any other type of engine. As such, to keep with the spirit of an ABC as an abstract base class that defines what methods a non-abstract class inheriting from it should implement, maybe some of the methods implemented here would be better off existing in the PatchPredictor(EngineABC) class, rather than within the ABC itself?

shaneahmed:
> As a general comment, I feel like the EngineABC does 'too much'. Some of the implemented methods are specific to a PatchPrediction engine, and would not be relevant for any other type of engine. As such, to keep with the spirit of an ABC as an abstract base class that defines what methods a non-abstract class inheriting from it should implement, maybe some of the methods implemented here would be better off existing in the PatchPredictor(EngineABC) class, rather than within the ABC itself?
> 
> Also, I think some of the examples do not make sense/wouldn't work in the documentation, so need to be double-checked

Thanks @measty. This needs to be fixed. As highlighted we need to check the examples as well before merge.

We will create a separate PR for PatchPredictor and keep only the code which is relevant to Engine"ABC".

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -396,19 +423,126 @@ def setup_patch_dataset(
...
+        return_labels = kwargs["return_labels"] if "return_labels" in kwargs else False
+
+        for _, batch_data in enumerate(dataloader):
+            batch_output_probabilities = self.model.infer_batch(
+                self.model,
+                batch_data["image"],
+                device=self.device,
+            )
+            # We get the index of the class with the maximum probability
+            batch_output_predictions = self.model.postproc_func(
```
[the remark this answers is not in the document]
mostafajahanifar: I'm just thinking out load here, but maybe a model-specific DataHandler class can be designed that can take care of both input preprocessing and output post-processing. And, the only things that engines do will be coordinate between the DataHandler and how results should be accumulated to form Tile or WSI-level results. 
Looking into HuggingFace Tokenizer and Pipeline classes can give us some good design clues.

mostafajahanifar (commented):
Thank you very much for putting together this PR, which looks OK to me. However, it does also look very much like what we had but just a little bit more tidy. Reviewing this PR made me think how much the new engine would make the life of users easier or our Toolbox more interesting. 
There are a few things that I would like to consider in out model and engine design before we go forward. I think I have made my opinion clear about the changes I have in mind However, these need serious thinking and discussion from the group because making those changes maybe not too good or appropriate for all the cases or very costly. Happy to discuss it further in a lab meeting :)

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -51,92 +63,193 @@ def prepare_engines_save_dir(
...
             <https://tia-toolbox.readthedocs.io/en/latest/pretrained.html>`_
             By default, the corresponding pretrained weights will also
             be downloaded. However, you can override with your own set
             of weights via the `weights` argument. Argument
             is case-insensitive.
         ioconfig (ModelIOConfigABC):
-            Input IO configuration to run the Engine.
-        _ioconfig ():
+            Input IO configuration of type :class:`ModelIOConfigABC` to run the Engine.
+        _ioconfig (ModelIOConfigABC):
```
measty, line 245: Do we really need two different ioconfigs? I think it is a source of potential confusion to users. If we really need both and can't find a way to have just one, we at least should provide more info about what distinguishes them, why they both are needed and how they interact. For example I think a user might wonder things like, if I update ioconfig will runtime _ioconfig automatically update to match?

tiatoolbox/utils/misc.py
```diff
@@ -1214,7 +1240,7 @@ def dict_to_store(
         patch_output (dict):
             A dictionary in the TIAToolbox Engines output format. Important
```
measty, line 1241: 'the TIAToolbox Engines output format' - what does this mean? This function works specifically for output of PatchPredictor. The engine in general can output things in a variety of formats depending what sort of engine it is (Patch Predictor, SemanticSegmentor etc output will look quite different from each other).

measty (commented):
Loos good so far, just added a few comments.

It will be easier to give feedback and see any issues that might need to be addressed once we start trying to implement the other engines we have using this ABC.

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -297,119 +510,250 @@ def pre_process_patches(
...
-                raw_predictions["labels"].extend(list(batch_data["label"]))
+                raw_predictions = self._update_model_output(
+                    raw_predictions=raw_predictions,
+                    key="labels",
+                    raw_output=batch_data["label"].numpy(),
+                )
+            if self.cache_mode:
+                zarr_group = write_to_zarr_in_cache_mode(
+                    zarr_group=zarr_group, output_data_to_save=raw_predictions
+                )
```
shaneahmed, line 589: raw_predictions should be set to None after updating zarr.

tiatoolbox/utils/misc.py
```diff
@@ -1202,9 +1202,35 @@ def add_from_dat(
...
+def patch_predictions_as_annotations(
+    preds: list,
+    keys: list,
+    class_dict: dict,
+    class_probs: list,
+    patch_coords: list,
+    classes_predicted: list,
+    labels: list,
+) -> list:
+    """Helper function to generate annotation per patch predictions."""
```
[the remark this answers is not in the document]
shaneahmed: This needs to be added. We should update this when we add WSI predictions in #783

tiatoolbox/models/engine/engine_abc.py
shaneahmed: Addressed in https://github.com/TissueImageAnalytics/tiatoolbox/pull/737/commits/af641e392370f1ada351d0811cae949f4b30e284

tiatoolbox/models/engine/engine_abc.py
shaneahmed: Hovernet requires these two ioconfigs. We can revisit this when we finalise engines for segmentation models.

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -155,43 +268,84 @@ class EngineABC(ABC):
...
+        resolution (Resolution):
+            Resolution used for reading the image. Please see
+            :class:`WSIReader` for details.
+            When `patch_mode` is True, the input image patches are expected to be at
+            the correct resolution and units. When `patch_mode` is False, the patches
+            are extracted at the requested resolution and units. Default value is 1.0.
+        units (Units):
+            Units of resolution used for reading the image. Choose
+            from either `level`, `power` or `mpp`. Please see
+            :class:`WSIReader` for details.
```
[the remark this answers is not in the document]
shaneahmed: Added `baseline` options.

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -1,47 +1,59 @@
...
     from tiatoolbox.annotation import AnnotationStore
+    from tiatoolbox.models.models_abc import ModelABC
+    from tiatoolbox.typing import IntPair, Resolution, Units
     from tiatoolbox.wsicore.wsireader import WSIReader
 
-    from .io_config import ModelIOConfigABC
-
 
 def prepare_engines_save_dir(
     save_dir: os | Path | None,
```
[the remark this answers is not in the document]
shaneahmed: It's not required when patch_mode is true and cachemode is false.

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -155,43 +268,84 @@ class EngineABC(ABC):
...
         batch_size (int):
             Number of images fed into the model each time.
+        cache_mode (bool):
+            Whether to run the Engine in cache_mode. For large datasets,
+            we recommend to set this to True to avoid out of memory errors.
+            For smaller datasets, the cache_mode is set to False as
+            the results can be saved in memory. cache_mode is always True when
+            processing WSIs i.e., when `patch_mode` is False. Default value is False.
+        cache_size (int):
+            Specifies how many images patches to process in a batch when
```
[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/737/commits/3774dcdef4c8a34050ca7e26f324465b64ba8c9d

tiatoolbox/utils/misc.py
```diff
@@ -1325,3 +1354,142 @@ def dict_to_zarr(
...
+        zarr_group (zarr.group):
+            Zarr group name consisting of zarr(s) to save the batch output
+            values.
+        output_data_to_save (dict):
+            Output data from the Engine to save to Zarr.
+        **kwargs (dict):
+            Keyword Args to update zarr_group attributes.
+
+    Returns:
+        Path to zarr file storing the patch predictor output
```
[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/737/commits/3774dcdef4c8a34050ca7e26f324465b64ba8c9d

tiatoolbox/utils/misc.py
```diff
@@ -1325,3 +1354,142 @@ def dict_to_zarr(
...
+            Coordinates batch output from infer wsi.
+        batch_output_label (np.ndarray):
+            Labels batch output from infer wsi.
+        save_path (str or Path):
+            Path to save the zarr file.
+        **kwargs (dict):
+            Keyword Args to update wsi_batch_output_to_zarr_group attributes.
+
+    Returns:
+        Path to zarr file storing the patch predictor output
```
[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/737/commits/3774dcdef4c8a34050ca7e26f324465b64ba8c9d

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -1,47 +1,59 @@
...
-    from .io_config import ModelIOConfigABC
-
 
 def prepare_engines_save_dir(
     save_dir: os | Path | None,
-    len_images: int,
     *,
     patch_mode: bool,
-    overwrite: bool,
+    overwrite: bool = False,
```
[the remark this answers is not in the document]
shaneahmed: The behaviour is the same as `mkdir`.

mostafajahanifar (approved):
All looks good to me 💯 

Well done on this PR 👍

Issue #610, opened 2023-05-19
Issue in patch extraction from images with more than 3 channels


Arwa-AlRubaian:
- TIA Toolbox version:1.3.3
- Python version: 3.10
- Operating System: Ubuntu

### Description
I was using (patchextractor=patchextraction.SlidingWindowPatchExtractor) to extract patches on images with 6 channels. Then when iterating over the patches using : 
for patch in patchextractor:
    #(some code ...)

the following erroroccur: 
Traceback (most recent call last):
    for patch in patchextractor:
  File "/****/site-packages/tiatoolbox/tools/patchextraction.py", line 167, in __next__
    return self[n]
  File "/****/site-packages/tiatoolbox/tools/patchextraction.py", line 179, in __getitem__
    return self.wsi.read_rect(
  File "/****/site-packages/tiatoolbox/wsicore/wsireader.py", line 2846, in read_rect
    return self._read_rect_at_resolution(
  File "/****/site-packages/tiatoolbox/wsicore/wsireader.py", line 970, in _read_rect_at_resolution
    return self.read_bounds(
  File "/****//site-packages/tiatoolbox/wsicore/wsireader.py", line 3057, in read_bounds
    return utils.transforms.background_composite(image=im_region)
  File "/****//site-packages/tiatoolbox/utils/transforms.py", line 41, in background_composite
    image = Image.fromarray(image)
  File "/****/site-packages/PIL/Image.py", line 3083, in fromarray
    raise TypeError(msg) from e
TypeError: Cannot handle this data type: (1, 1, 6), |u1

John-P:
We do not currently have support in place for multi-channel images as we have been focused on RGB bright-field imaging. However, we would love to add support for this and getting patch extraction working for multi-channel images would be a good start. I think the trickiest part would be to define how patches get written out to disk. Do you have a preferred format for this? Would a multi-channel TIFF or npy file be suitable for your needs?

rogerdds:
> 

Maybe this should be left up to the user? as long as the patch_extractor returns a np array e.g. H,W,num_chans it should be good?

rogertrullo:
Maybe this should be left up to the user? as long as the patch_extractor returns a np array e.g. H,W,num_chans it should be good?

PR #742, opened 2023-11-22
:children_crossing: Handle Multidimensional `Features` in `WSIReader`

### LINKED ISSUE #609: :monocle_face: Rename `VirtualWSIReader` Modes
- TIA Toolbox version: 1.4.0

### Description
Currently, `VirtualWSIReader` [supports](https://github.com/TissueImageAnalytics/tiatoolbox/blob/f684b863f4d309d8c2da0ae93f0ae4243e8ba650/tiatoolbox/wsicore/wsireader.py#L2597) two modes: `rgb` and `bool`:

- However, in reality, `bool` mode can process single-channel `np.uint8` images. Even now, without code modifications, it works fine on matrixes with values other than `0` or `1`. `uint8` is wider than `bool` but almost always takes the same amount of memory. 
- If someone uses it with **HSV** images, it will work just fine, except for the fact that VirtualWSIReader will be called with `mode="rgb"`, which makes no sence. 

Let's rename these modes to `single channel` and `multichannel`?

behnazelhaminia:
- Adds support for multichannel inputs e.g., image masks, feature maps etc. Fixes #609 and #610


tiatoolbox/utils/transforms.py
```diff
@@ -42,6 +42,9 @@ def background_composite(
         >>> plt.show()
 
     """
+    if image.size[:, :, 0] >= 4:
```
shaneahmed, line 45: This will fail for RGBA image. I think the best way to handle this is using mode https://github.com/TissueImageAnalytics/tiatoolbox/issues/609

test.py
```diff
@@ -0,0 +1,32 @@
+"""Import modules required to run the Jupyter notebook."""
```
shaneahmed, line 1: As this test should be related to WSIReader. Please add a test in `https://github.com/TissueImageAnalytics/tiatoolbox/blob/develop/tests/test_wsireader.py`

shaneahmed: Please check the tests in this file which will give you an idea on how to write these tests.

tiatoolbox/wsicore/wsireader.py
```diff
@@ -2809,15 +2818,22 @@ def __init__(
...
+
         if isinstance(input_img, np.ndarray):
             self.img = input_img
         else:
             self.img = utils.imread(self.input_path)
 
+        if mode != "bool" and (
+            self.img.ndim == 2 or self.img.shape[2] not in [3, 4]  # noqa: PLR2004
+        ):
+            mode = "feature"
```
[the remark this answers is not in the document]
shaneahmed: Logging would be better option here as background composite would fail in case the input is not rgb or rgba. Also, node should be bool for a boolean input.

tiatoolbox/wsicore/wsireader.py
```diff
@@ -3267,6 +3283,9 @@ class docstrings for more information.
         if interpolation in [None, "none"]:
             interpolation = None
 
+        if interpolation == "optimise" and self.mode == "bool":
```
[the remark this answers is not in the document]
shaneahmed: For features, it can be chosen by the user what to select as it depends on the scenario.
For binary masks, it should be bool otherwise it will generate unintended output.

tests/test_wsireader.py
```diff
@@ -9,6 +9,10 @@
 from copy import deepcopy
 from pathlib import Path
 
+import os
+
+os.add_dll_directory("C:/Users/u2272723/PycharmProjects/openslide-win64/bin")
+
```
shaneahmed: Please delete these lines.

tiatoolbox/wsicore/wsireader.py
```diff
@@ -2830,6 +2830,7 @@ def __init__(
         if mode != "bool" and (
             self.img.ndim == 2 or self.img.shape[2] not in [3, 4]  # noqa: PLR2004
         ):
+            logger.warning("The input image mode is set to 'feature'")
```
shaneahmed, line 2833:
```suggestion
            logger.warning("The image mode is set to "feature" as the input dimensions do not match with binary mask or RGB/RGBA.")
```

shaneahmed (approved):
Thanks @behnazelhaminia

PR #744, opened 2023-11-23
:hammer: Add `utils/misc.py` to mypy

Jiaqi-Lv:
- Add `utils/misc.py` to mypy checks

### DIFF
--- a/tiatoolbox/utils/misc.py
+++ b/tiatoolbox/utils/misc.py
@@ -36,12 +36,12 @@ if TYPE_CHECKING:  # pragma: no cover
 
 
 def split_path_name_ext(
-    full_path: PathLike,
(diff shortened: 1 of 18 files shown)


tiatoolbox/utils/misc.py
```diff
@@ -154,18 +153,18 @@ def imwrite(image_path: os | PathLike, img: np.ndarray) -> None:
 
     """
     if isinstance(image_path, Path):
-        image_path = str(image_path)
+        image_path_str = str(image_path)
```
shaneahmed, line 156: what if image_path is not a Path?

I think you can remove the if statement here and just run str without check.

shaneahmed (approved):
Thanks @Jiaqi-Lv

Issue #722, opened 2023-10-16
WSIReader Tile Sizes Issue - SVS Format


TomastpPereira:
- TIA Toolbox version: 1.4.1
- Python version: 3.11.3
- Operating System: Win 10 x64

### Description

I have been attempting to create WSI tiles using the WSIReader's save_tiles function but have been getting tiles of incorrect shape and content when using WSIs in svs format. 

For a set of WSIs in svs format, the output set of tiles does not match the specified output size at certain objective values. 
By this I mean that I have passed for example, tile_objective_value=5 and tile_read_size=(1024,1024) and the result is a set of patches at (256,256). These tiles also seem to only contain white space. A requested output of (512,512) also instead created (128,128). 
When testing the same issue on sample WSIs provided by Openslide, the issue persists. 

This issue also does not seem to persist when using a WSI in ndpi format. 
The output when using the CMU-1 sample slide from OpenSlide outputs 132 items when using the ndpi format, and only 9 items of incorrect shape when using the svs format. 

I have noted that this seems to occur for all objective values which are below the highest available (20x for the sample).  

### What I Did

```
import openslide
import tiatoolbox
from tiatoolbox.wsicore.wsireader import WSIReader

wsi = WSIReader.open(input_img="C:\\...\\Downloads\\CMU-1.svs")
wsi.save_tiles(output_dir="...", tile_objective_value=5, tile_read_size=(1024,1024))
```

shaneahmed:
Thanks @TomastpPereira Please can you output the type of WSIReader object both for svs and ndpi case? We need to make sure that the objects are created correctly.

TomastpPereira:
Hello @shaneahmed,
For both cases, the type is <class 'tiatoolbox.wsicore.wsireader.OpenSlideWSIReader'>.

In the cases of the svs type, I receive the warning below:
[WARNING] Read: Scale > 1.This means that the desired resolution is higher than the WSI baseline (maximum encoded resolution). Interpolation of read regions may occur.
However, this warning also appears when passing a higher objective value (20), in which case the issue does not occur.  


I have additionally attached the output files for the slide in both types. 
[Output - CMU-1.ndpi.csv](https://github.com/TissueImageAnalytics/tiatoolbox/files/12929895/Output.-.CMU-1.ndpi.csv)
[Output - CMU-1.svs.csv](https://github.com/TissueImageAnalytics/tiatoolbox/files/12929893/Output.-.CMU-1.svs.csv)

[...]

PR #752, opened 2023-12-08
🩹 Fix `save_tiles`

measty:
- Fix issue #722


measty:
> I ran more tests on 8 WSIs from OpenSlide samples and 5 from the ACROBAT dataset and I don't see any issues, except that the size of tiles becomes inconsistent at the edges of the WSI, e.g., instead of (1024, 1024) for a tile, you can get (x, 1024) or (x, y) with x and y being integer numbers less than 1024.
> 
> Here are some CSVs that showcase that (scroll to the bottom): [CMU-3.ndpi-5x.csv](https://github.com/TissueImageAnalytics/tiatoolbox/files/13648598/CMU-3.ndpi-5x.csv) [CMU-3.svs-20x.csv](https://github.com/TissueImageAnalytics/tiatoolbox/files/13648599/CMU-3.svs-20x.csv) [JP2K-33003-1.svs-10x.csv](https://github.com/TissueImageAnalytics/tiatoolbox/files/13648601/JP2K-33003-1.svs-10x.csv)
> 
> If that's expected behaviour, then feel free to close this PR.

As far as I know, that is the expected behaviour. Certainly that is how it behaved before on the slides where it was working, so in this PR I have not changed that, I have just fixed the behaviour on the slides where it was returning nonesense tiles at wrong resolutions.

If it is wanted that we change the behavior to returning padded tiles near edges, it is a simple change; we just have to remove the lines:

```
            end_h = min(end_h, slide_h)
            end_w = min(end_w, slide_w)
```

Issue #692, opened 2023-08-11
On Windows tests raise an error `sqlite3.OperationalError: no such function: POWER`


shaneahmed:
- TIA Toolbox version: 1.4.1 (develop)
- Python version: 3.9
- Operating System: Windows

### Description
Run test `test_nquery_boxpoint_boxpoint`

https://github.com/TissueImageAnalytics/tiatoolbox/blob/2da75c6b7b3bda543dcfa14f2d81f20431094464/tests/test_annotation_stores.py#L1704
On Windows it raises an error. It passes on Linux.

```
..\tiatoolbox\annotation\storage.py:1568: in nquery
    subquery_result = self.query(
..\tiatoolbox\annotation\storage.py:2768: in query
    cur = self._query(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <tiatoolbox.annotation.storage.SQLiteStore object at 0x000001F2DF10F6D0>
columns = '[key], properties, cx, cy, geometry', geometry = <POINT (1 1)>
callable_columns = '[key], properties, cx, cy, geometry'
geometry_predicate = 'centers_within_k', where = "props['class'] != 'A'"
min_area = None, distance = 2

    def _query(
        self,
        columns: str,
        geometry: Geometry | None = None,
        callable_columns: str | None = None,
        geometry_predicate="intersects",
        where: Predicate | None = None,
        min_area: float | None = None,
        distance: float = 0,
        *,
        unique: bool = False,
        no_constraints_ok: bool = False,
        index_warning: bool = False,
    ) -> sqlite3.Cursor:
        """Common query construction logic for `query` and `iquery`.
    
        Args:
            columns(str):
                The columns to select.
            geometry(tuple or Geometry):
                The geometry being queried against.
            callable_columns(str):
                The columns to select when a callable is given to
                `where`.
            geometry_predicate(str):
                A string which define which binary geometry predicate to
                use when comparing the query geometry and a geometry in
                the store. Only annotations for which this binary
                predicate is true will be returned. Defaults to
                "intersects". For more information see the `shapely
                documentation on binary predicates <https://shapely.
                readthedocs.io/en/stable/manual.html#binary-predicates>`_.
            where (str or bytes or Callable):
                The predicate to evaluate against candidate properties
                during the query.
            unique(bool):
[...]

PR #769, opened 2024-01-12
♻️ Remove `POWER` from sqlite-dist Calculations

measty:
Removes the use of POWER from distance calculations to address issue #692

Replaces use of POWER with explicit multiplication.


shaneahmed:
This passes on Windows now. Thanks @measty

shaneahmed (approved):

Issue #771, opened 2024-01-17
Unexpected behaviour in annotation store when property key has a . in it


measty:
- TIA Toolbox version: All
- Python version: All
- Operating System: All

### Description

If a properties dictionary of an annotation has a . in it, some operations of AnnotationStore behave unexpectedly if the operation involves the usage of json_extract

### Cause

The root of the issue is in the line:

https://github.com/TissueImageAnalytics/tiatoolbox/blob/3cea66215e0f340534d5a6b2b34b89dcabd70939/tiatoolbox/annotation/dsl.py#L268

In the case of a key which contains a . this ends up being incorrectly parsed. For example, if properties dict has a key 'MSI.status' and you try to do an operation which needs to access that property from the json, it gets parsed into a json_extract function call that instead of accessing a field 'MSI.status' in the JSON, tries to access a subfield 'status' of a field 'MSI' in the JSON

PR #772, opened 2024-01-20
🐛 Fix `json_extract` Construction

measty:
Restructures the way the dsl builds the SQL expression to access JSON properties so that properties with a . in are correctly treated as a single key

Adds a test to check for correct behaviour when accessing a property containing a .

Addresses issue #771 

Would like to check with John if removing json.dumps() is ok. I couldn't see a reason it was necessary and all tests pass/all the use-cases I tried worked fine, but maybe there is one I missed.


shaneahmed (approved):
Thanks @measty

PR #776, opened 2024-01-26
⚡️ Add `torch.compile` to `PatchPredictor`

Abdol:
This mini-PR adds `torch.compile` functionality to `PatchPredictor`.

### DIFF
--- a/tiatoolbox/models/architecture/utils.py
+++ b/tiatoolbox/models/architecture/utils.py
@@ -39,7 +39,7 @@ def compile_model(
         return model
 
     # This check will be removed when torch.compile is supported in Python 3.12+
-    if sys.version_info >= (3, 12):
+    if sys.version_info >= (3, 12):  # pragma: no cover
         logger.warning(
             ("torch-compile is currently not supported in Python 3.12+. ",),
         )


measty:
At the moment the model is compiled before it is sent to the GPU (if GPU is being used). I think at least some of what torch.compile does is device-aware, so it may be better to compile after it is sent to device. Have you tried testing if the ordering makes a difference?

shaneahmed:
> At the moment the model is compiled before it is sent to the GPU (if GPU is being used). I think at least some of what torch.compile does is device-aware, so it may be better to compile after it is sent to device. Have you tried testing if the ordering makes a difference?



> At the moment the model is compiled before it is sent to the GPU (if GPU is being used). I think at least some of what torch.compile does is device-aware, so it may be better to compile after it is sent to device. Have you tried testing if the ordering makes a difference?

Agree, this should be checked.

tiatoolbox/models/architecture/utils.py
```diff
@@ -2,10 +2,60 @@
...
+    # This check will be removed when torch.compile is supported in Python 3.12+
+    if sys.version_info >= (3, 12):
+        logger.warning(
+            ("torch-compile is currently not supported in Python 3.12+. ",),
+        )
+        return model
+
+    if isinstance(
+        model,
+        torch._dynamo.eval_frame.OptimizedModule,  # skipcq: PYL-W0212 # noqa: SLF001, E501
```
shaneahmed, line 50: Why E501?

tiatoolbox/models/architecture/utils.py
```diff
@@ -2,10 +2,60 @@
...
+    ):
+        logger.warning(
```
shaneahmed, line 52: Please can you fix coverage?

tiatoolbox/models/architecture/utils.py
```diff
@@ -2,10 +2,60 @@
...
+            ("The model is already compiled. ",),
+        )
+        return model
```
shaneahmed, line 55: Please can you fix coverage?

tiatoolbox/models/architecture/utils.py
```diff
@@ -2,10 +2,60 @@
...
+            `max-autotune-no-cudagraphs`.
+        disable (bool):
+            If True, torch-compile will be disabled.
+
+    Returns:
+        Callable:
+            Compiled model.
+
+    """
+    if disable:
```
shaneahmed, line 38: I do not think we need this variable. I think we should only call this function if `not disable`

tiatoolbox/models/engine/patch_predictor.py
```diff
@@ -250,7 +251,13 @@ def __init__(
 
         self.ioconfig = ioconfig  # for storing original
         self._ioconfig = None  # for storing runtime
-        self.model = model
+        self.model = (
+            compile_model(  # for runtime, such as after wrapping with nn.DataParallel
+                model,
+                mode=rcParam["torch_compile_mode"],
```
shaneahmed, line 257: Do we need rcparam for this? We can just set this as kwargs argument in the engines.

PR #777, opened 2024-01-29
🐛 Fix `TIFFWSIReader` `read_bound`

vqdang:
image

Emergency bugfix per @John-P request.
The culprit is reading bound doesn't use the adjusted bounds as have been done in OpenSlideReader.


shaneahmed:
Thanks @vqdang  and @John-P Please can you add a simple test to make sure this produces expected results.

measty:
I think this also needs a fix along similar lines for read_rect

measty:
I've added a few changes  to this PR too.
 - fix for read_rect along similar lines to read_bounds (it was bugged in the same way)
 - a few changes to make the reader compatible with a wider variety of tiffs
 - a change to metadata in wsireader that fixes some performance issues on some wsis

shaneahmed (approved):
Thanks @Abdol , @measty  and @vqdang for your help in fixing this bug.

PR #783, opened 2024-02-09
🧑‍💻 Define `PatchPredictor`

shaneahmed:
- Redesigns PatchPredictor engine using the new EngineABC base class.
- The WSIs are now processed using the same code as for the processing the patches using WSI based dataloader.
- The intermediate output is saved as zarr for the WSIs to resolve memory issues.
- The output of model architectures should now be a dictionary.
- The output can be specified as AnnotationStore for visualisation using TIAViz.


shaneahmed:
Thanks Abdullah. However, we will merge #737 first before working on this PR.

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -558,6 +561,9 @@ def infer_patches(
         if self.return_labels:
             keys.append("labels")
 
+        if return_coordinates:
```
shaneahmed, line 564:
```suggestion
        if return_coordinates or self.output_type == "AnnotationStore":
```

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -571,9 +577,14 @@ def infer_patches(
                 batch_data["image"],
                 device=self.device,
             )
+            if return_coordinates:
+                batch_output["coordinates"] = batch_data["coords"].numpy()
 
             if self.return_labels:  # be careful of `s`
-                batch_output["labels"] = batch_data["label"].numpy()
+                if isinstance(batch_data["label"], torch.Tensor):
```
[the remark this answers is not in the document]
shaneahmed: This change will turn into a very long line.

tiatoolbox/models/engine/patch_predictor.py
```diff
@@ -208,526 +279,143 @@ class PatchPredictor(EngineABC):
...
+        self: EngineABC,
+        raw_predictions: dict | Path,
+        **kwargs: Unpack[EngineABCRunParams],
+    ) -> dict | Path:
+        """Post process WSI output.
 
-        Returns:
-            :class:`numpy.ndarray`:
-                Model predictions of the input dataset
+        Takes the raw output from patch predictions and post-process to improve the
```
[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/783/commits/0d1dda925a94623b9e31a74c6b2844633c0160e0

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -535,6 +535,9 @@ def infer_patches(
                 An :class:`torch.utils.data.DataLoader` object to run inference.
             save_path (Path | None):
                 If `cache_mode` is True then path to save zarr file must be provided.
+            return_coordinates (bool):
+                Whether to save coordinates in the output. This is required when
+                this function is called by `infer_wsi` and the `patch_mode` is False.
```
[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/783/commits/0d1dda925a94623b9e31a74c6b2844633c0160e0

tiatoolbox/models/engine/patch_predictor.py
```diff
@@ -183,17 +254,17 @@ class PatchPredictor(EngineABC):
         >>> # list of 2 image patch files as input
         >>> data = ['path/img.png', 'path/img.png']
         >>> predictor = PatchPredictor(pretrained_model="resnet18-kather100k")
```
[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/783/commits/0d1dda925a94623b9e31a74c6b2844633c0160e0

tiatoolbox/models/engine/patch_predictor.py
```diff
@@ -183,17 +254,17 @@ class PatchPredictor(EngineABC):
...
-        >>> output = predictor.predict(data, mode='patch')
+        >>> output = predictor.run(data, mode='patch')
 
         >>> # list of 2 image tile files as input
         >>> tile_file = ['path/tile1.png', 'path/tile2.png']
         >>> predictor = PatchPredictor(pretraind_model="resnet18-kather100k")
```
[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/783/commits/0d1dda925a94623b9e31a74c6b2844633c0160e0

tiatoolbox/models/engine/patch_predictor.py
```diff
@@ -183,17 +254,17 @@ class PatchPredictor(EngineABC):
...
-        >>> output = predictor.predict(tile_file, mode='tile')
+        >>> output = predictor.run(tile_file, mode='tile')
 
         >>> # list of 2 wsi files as input
         >>> wsi_file = ['path/wsi1.svs', 'path/wsi2.svs']
         >>> predictor = PatchPredictor(pretraind_model="resnet18-kather100k")
```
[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/783/commits/0d1dda925a94623b9e31a74c6b2844633c0160e0

tiatoolbox/utils/misc.py
```diff
@@ -1203,10 +1203,10 @@ def add_from_dat(
 
 
 def patch_predictions_as_annotations(
-    preds: list,
+    preds: list | np.ndarray,
```
measty, line 1206: If preds can be empty list this function wont work. Need to iterate over coords instead which always must be there, and save preds only if they exist. At the moment using patch predictor on a WSI with save_type="AnnotationStore" just saves an empty store

measty (changes requested):
I have done an initial review and commented a few things that need changing.

A few other notes as well: this is awkward to review as some changes are wrapped up in the branch this merges into - some of the issues I have found don't show up in the code changed for this PR cause the actual change is in the dev-update-engine-abc branch.

There will need to be a review of the notebooks before this gets merged, as this will at the very least break the patch-prediction notebook and perhaps others.

Also, is the change of behaviour for a model like "resnet18-kather100k" intentional? previously it would return hard class predictions also, now it does not

tiatoolbox/models/engine/patch_predictor.py
```diff
@@ -117,63 +110,141 @@ class PatchPredictor(EngineABC):
...
-        model (nn.Module):
-            Use externally defined PyTorch model for prediction with.
-            weights already loaded. Default is `None`. If provided,
-            `pretrained_model` argument is ignored.
-        pretrained_model (str):
-            Name of the existing models support by tiatoolbox for
-            processing the data. For a full list of pretrained models,
-            refer to the `docs
+        model (str | ModelABC):
+            A PyTorch model. Default is `None`.
```
mostafajahanifar, line 114: would be better to update the docstring here to clarify that the user can pass a string to select from a model Zoo or provide a Pytorch model.

tiatoolbox/models/engine/engine_abc.py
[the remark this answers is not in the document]
shaneahmed: Fixed in https://github.com/TissueImageAnalytics/tiatoolbox/pull/783/commits/5085035af033698f97ac8feeeab438a892aff48d. Added tests to override. Need to add override warning.

tiatoolbox/models/engine/patch_predictor.py
shaneahmed: Fixed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/783/commits/45a339b53a2cecd8bc7c7c722024fa2e2fb0e6ed

tiatoolbox/utils/misc.py
measty: There is still some issues with how it is being saved.
1. the correct filename isn't being passed to engine.save_predictions, so the saved .db is just called the default output.db. It is being set in:
https://github.com/TissueImageAnalytics/tiatoolbox/blob/45a339b53a2cecd8bc7c7c722024fa2e2fb0e6ed/tiatoolbox/models/engine/engine_abc.py#L1041
but this is not being used anywhere.
2. The scale_factor is not being set and passed to save_predictions, so the patches are currently saved and can be visualized, but appear wrongly scaled. The scale factor needs to be calculated as model_mpp (from io config) / slide_mpp somewhere in the pipeline and added to kwargs.

shaneahmed: Addressed (1) in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/783/commits/20362efe7097dd8ad19d0b9772921fed03387f7d

shaneahmed: Addressed (2) in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/783/commits/da0ce4fec1a9d9d098d95c9c398cd6cf583ade40. I am still not sure if this is correct as the output does not match. The scale_factor is being passed to `dict_to_store`. However, couldn't verify in the visualization tool. Also, all the output is single colour, need to verify this as well.

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -982,13 +1020,89 @@ def _run_patch_mode(
...
+                slide_resolution = wsimeta_dict[dataloader_units]
+                scale_factor = tuple(np.divide(slide_resolution, dataloader_resolution))
+
+            if dataloader_units == "level":
+                wsimeta_dict = dataloader.dataset.reader.info.as_dict()
+                downsample_ratio = wsimeta_dict["level_downsamples"][
+                    dataloader_resolution
+                ]
+                scale_factor = (1.0 / downsample_ratio, 1.0 / downsample_ratio)
+
```
shaneahmed, line 1084: Add `objective_power`

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -982,13 +1020,89 @@ def _run_patch_mode(
...
+                masks=mask,
+                patch_mode=False,
+                ioconfig=self._ioconfig,
+            )
+
+            # get units and resolution from dataloader.
+            dataloader_units = dataloader.dataset.units
+            dataloader_resolution = dataloader.dataset.resolution
+
+            # if dataloader units is baseline slide resolution is 1.0.
```
shaneahmed, line 1068: move scaling to helper function.

shaneahmed:
@mostafajahanifar and @adamshephard I have pushed the updates as discussed during the meeting.

tiatoolbox/models/engine/patch_predictor.py
```diff
@@ -117,63 +110,141 @@ class PatchPredictor(EngineABC):
...
+            are extracted at the requested resolution and units.
+            Default value is `baseline`.
         verbose (bool):
-            Whether to output logging information.
+            Whether to output logging information. Default value is False.
 
     Examples:
         >>> # list of 2 image patches as input
         >>> data = ['path/img.svs', 'path/img.svs']
         >>> predictor = PatchPredictor(pretrained_model="resnet18-kather100k")
```
mostafajahanifar, line 246: `pretrained_model` should be replaced with `model`

mostafajahanifar: the same goes for the next example

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -571,9 +578,14 @@ def infer_patches(
                 batch_data["image"],
```
mostafajahanifar, line 578: This is where I have a problem with the design! (I guess I mentioned this when we doing ModelABC).

I think we should get rid of ModelABC altogether, if possible! Instead of `self.model.infer_batch`, we should call `self.infer_batch`, in other words, `infer_batch` should be a method of the engine, not the model! The same goes with the `preproc_func` and `postproc_func`. These are static methods anyway. 

The reason I am suggesting this is to make the engines more versatile and the whole toolbox easier to use with new models. I understand that we may end up needing new modules for loading preprocessing, postprocessing, and weight loading of the pretrained models (model zoo), but that also maybe make it more organized and help avoid repetition. If we are going to do this, it is better to be done early rather than late.

mostafajahanifar (changes requested):
Thank you for your excellent work on this PR, Shan. The extent of your contributions is truly impressive.

While the current design appears functional, I believe we can enhance the flexibility of the toolbox by addressing the use of ModelABC for models. To achieve this, I suggest relocating methods that specifically pertain to ModelABC to the engines. This will allow for greater customization and the ability to seamlessly swap engines and models based on user preferences. Predefined (default) functions for common model types can also be implemented to streamline the process.

Anyway, I understand that you may decide to include these changes in a separate PR.

tiatoolbox/models/engine/engine_abc.py
shaneahmed: We will need a separate PR to update this as discussed.

PR #785, opened 2024-02-14
📌 Update `wsidicom` dependency

shaneahmed:
- Update `wsidicom` dependency


shaneahmed:
Python 3.8 is only supported upto wsidicom version 0.13.0

shaneahmed:
This PR is dependent on #786 .

PR #789, opened 2024-02-21
🆕 Add Reading Slide-level info & Simple Plugins for Visualization Tool

measty:
This PR adds in a feature from the most recent version of TIAViz to allow a .csv to be provided in the slides folder that can contain slide level predictions or ground truth labels/other slide level info.

When loading a slide in the UI, if the slide name (including suffix) appears in the "Image File" column of the .csv, any information in the corresponding row of the .csv will appear in the UI below the main window in a nicely formatted string.

Only the "Image File" column is required for this to work, the rest of the columns can be anything you need to have displayed.

Also adds entry in documentation describing feature.

In response to Mostafa's request for more flexibility in what sort of extra info you can add to the UI, i've also added some basic plugin functionality to this PR (and converted to draft for now as it has become much larger and needs tests/docs to be added over the coming weeks, and may be tweaked a bit).

Plugins are defined as a class following an ABC and can add a collection of bokeh UI elements and/or plots below the main view window (and can potentially also insert additional UI elements into the main controls section). This is pretty flexible and can add some quite complex extras in there as you have all the power of bokeh to work with.

I have added 3 pre-built plugins that will cover some use-cases, and provide examples which users can adapt or take inspiration from to make their own custom plugin.

The 3 provided are:
ImageGrid: if each slide has a folder of images associated with it, this plugin will show the images in a grid-like layout below the slide view window.
BarPLotGrid: makes a gridlike layout of bar charts from csv files associated with each slide
StatsPlot: adds some UI elements to show a histogram showing the distribution of a property over the annotations in the selected box on the slide window

To add plugins when launching the tool, you can add one or more --plugin arguments  each with a path pointing to the file where the plugin class is.


tiatoolbox/visualization/bokeh_app/main.py
```diff
@@ -2102,6 +2170,7 @@ def setup_doc(self: DocConfig, base_doc: Document) -> tuple[Row, Tabs]:
         base_doc.add_root(control_tabs)
         base_doc.add_root(popup_table)
         base_doc.add_root(slide_info)
+        base_doc.add_root(extra_layout)
         base_doc.title = "Tiatoolbox Visualization Tool"
```
shaneahmed, line 2174:
```suggestion
        base_doc.title = "TIAToolbox Visualization Tool"
```

tiatoolbox/visualization/bokeh_app/templates/__init__.py
```diff
@@ -0,0 +1 @@
+"""plugins and templates for bokeh visualization app."""
```
shaneahmed, line 1:
```suggestion
"""Plugins and templates for bokeh visualization app."""
```

tiatoolbox/visualization/bokeh_app/templates/bar_plot_grid.py
```diff
@@ -0,0 +1,100 @@
...
+
+        Args:
+            slide_path (str): The path to the slide for which the extra layout is
+            being created
+            UI: contains the main UI elements of the bokeh app
+            old_children: contains the previous children of the layout
+
+        Returns:
+            list: A list containing the new children of the extra layout
+        """
```
shaneahmed, line 35:
```suggestion

        """
```

tiatoolbox/visualization/bokeh_app/templates/stats_plot.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Args:
+            slide_path (str): The path to the slide for which the extra layout is being
+            created
+            UI: contains the main UI elements of the bokeh app
+            old_children: contains the previous children of the layout
+
+        Returns:
+            list: A list containing the new children of the extra layout
+        """
```
shaneahmed, line 98:
```suggestion

        """
```

tiatoolbox/visualization/ui_utils.py
```diff
@@ -29,3 +31,58 @@ def get_level_by_extent(extent: tuple[float, float, float, float]) -> int:
...
+    We provide a few pre-built plugins in tiatoolbox.visualization.bokeh_app.templates
+    folder. These can be used as a starting point for creating new plugins, or as-is if
+    they happen to meet your needs. Provided plugins are:
+    - ImageGrid: Displays a grid of images below the main view window from a specified
+    folder for each slide.
+    - BarPlotGrid: Displays a grid of bar plots below the main view window for each csv
+    file found within a folder for each slide.
+    - StatsPlot: Displays a histogram of the distributions of properties of annotations
+    contained within the selected box-shaped region of the slide.
+    """
```
shaneahmed, line 72:
```suggestion

    """
```

JiwaniZakir (commented):
In `tests/test_app_bokeh.py` around line 162, the `fetch_sample_to_dir` call for `"test_meta"` is missing its closing parenthesis and second argument (the destination directory), leaving the call syntactically broken — this will cause the fixture to fail at runtime. Compare it to every other `fetch_sample_to_dir` invocation in the same fixture, which all pass a destination path as the second positional argument.

The rename of `UI_elements_1`/`UI_elements_2` to `ui_elements_1`/`ui_elements_2` in `docs/visualization.rst` (lines 473 and 488) is a silently breaking change for anyone with an existing config file using the old key names. It would be worth either adding a deprecation path in the config-loading code that accepts both casings, or at minimum calling out the rename explicitly as a breaking change in the docs rather than quietly updating the example snippet.

The `--plugin` CLI help string in `tiatoolbox/cli/visualize.py` uses a Windows-style backslash path (`tiatoolbox\visualization\templates`) which will render as escape sequences or just look wrong on Linux/macOS — it should use forward slashes or `os.path.join`-style notation in documentation strings.

PR #796, opened 2024-03-15
:alien: Change `size` to `radius` for `bokeh 3.4` Compatibility

measty:
Bokeh 3.4 depreciated size in favor of radius in circle glyphs (you used to be able to use either, now only radius exists) and so tests now fail since 3.4.

This PR replaces all references to 'size' in circle glyphs with corresponding 'radius' for compatibility with bokeh 3.4.


shaneahmed (approved):
Thanks @measty

PR #801, opened 2024-03-21
Add `mypy` Type Check to `wsi_registration.py`

Jiaqi-Lv:
- Add `mypy` Type Check to `wsi_registration.py`
  - This adds `mypy` checks to all modules in `tools`.


tiatoolbox/tools/registration/wsi_registration.py
```diff
@@ -1656,6 +1667,8 @@ def read_rect(
             resolution=resolution,
             units=units,
         )
+        transformed_location: tuple[int, int]
```
shaneahmed, line 1670: I think this should not be required considering output of `self.get_transformed_location` is already specified.

Jiaqi-Lv: Yes that's true, I'll remove these

shaneahmed (approved):
Thanks @Jiaqi-Lv

PR #813, opened 2024-04-30
🐛 Fix Rendering Annotations with Holes

measty:
This PR adds correct rendering of annotations with holes.

Previously, annotations with holes would not be rendered. This PR fixes the issue.

image

Previously an overlay like the above, with a geometry with holes in it, would not display.


shaneahmed:
Thanks @measty Please can you add snapshots of comparison showing what was the previous problem and how it is fixed now in the description for this PR?

shaneahmed (approved):
Thanks @measty

PR #822, opened 2024-06-13
🆕 Use Annotations as a Filter for Patch Extraction

measty:
This PR is something i've been thinking about from a conversation with Daan when he was here from Radboud doing a seminar a while back. His comment was that one thing some of his colleges felt was missing in TIAtoolbox was ability to get patches filtered by class/label.

Its not that hard to set up tiatoolbox patch extractor to allow this, which is what i've added in this PR.

it allows you to provide a path to an annotation store as the mask. By default the mask will then be built out of all the annotations, but you can also provide any filter that works in the annotation store, and have only those annotations make the mask. So for example, if you have some pathologist annotations with some regions, and you only want to extract patches which have some overlap with regions labelled 'tumor' or whatever, you would just have to provide the appropriate string ("props['label'] == 'tumor'" for example).


tiatoolbox/tools/patchextraction.py
```diff
@@ -48,6 +49,7 @@ class ExtractorParams(TypedDict, total=False):
     input_mask: str | Path | np.ndarray | wsireader.VirtualWSIReader
```
shaneahmed, line 49: Update docstring for input_mask as it now accepts "AnnotationStore"

shaneahmed (approved):
Thanks @measty

shaneahmed:
> Thanks @measty

@mostafajahanifar Please can you review this PR so we can merge this.

shaneahmed:
@mostafajahanifar I have updated the file.

mostafajahanifar (approved):
Well done, Mark! Great PR 💯 

I only made some minor comments on the Notebook, feel free to make them or to ignore them.

PR #841, opened 2024-07-31
🆕 Improve Support for `QuPath` GeoJson and Multipoint Geometry

measty:
This PR makes a collection of changes related to reviewer comments in the TIAViz paper.

Main changes are:

1. Layer names in the interface for static image overlays now are named after filename stem, instead of generic 'layer1', 'layer2' etc to make it easier to distinguish the layers.
2. Adds multipoint geometry rendering as qupath exports collections of points as multipoints not individual point annotations
3. Robustifies qupath import from geojson to handle qupath special case of `nucleusGeometry`.

### DIFF
--- a/tests/test_tileserver.py
+++ b/tests/test_tileserver.py
@@ -100,11 +100,7 @@ def app(remote_sample: Callable, tmp_path: Path) -> TileServer:
     imwrite(thumb_path, thumb)
 
     sample_store = Path(remote_sample("annotation_store_svs_1"))
-    store = SQLiteStore(sample_store)
-    geo_path = tmp_path / "test.geojson"
-    store.to_geojson(geo_path)
-    store.commit()
-    store.close()
+    geo_path = Path(remote_sample("geojson_cmu_1"))
 
     # make tileserver with layers representing all the types
     # of things it should be able to handle
@@ -115,7 +111,7 @@ def app(remote_sample: Callable, tmp_path: Path) -> TileServer:
             "tile": str(thumb_path),
             "im_array": np.zeros(wsi.slide_dimensions(1.25, "power"), dtype=np.uint8).T,
             "overlay": str(sample_store),
-            "store_geojson": tmp_path / "test.geojson",
+            "store_geojson": str(geo_path),
         },
     )
     app.config.from_mapping({"TESTING": True})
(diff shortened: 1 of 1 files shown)


tests/test_app_bokeh.py
```diff
@@ -404,7 +404,8 @@ def test_load_img_overlay(doc: Document, data_path: pytest.TempPathFactory) -> N
     # trigger an event to select the image overlay
     click = MenuItemClick(layer_drop, str(data_path["img_overlay"]))
     layer_drop._trigger_event(click)
-    layer_slider = doc.get_model_by_name("layer2_slider")
+    l_name = "CMU-1-Small-Region_rendered_annotations"
```
shaneahmed: can this be changed to `{file_name}.stem` to avoid future errors?

tiatoolbox/utils/visualization.py
```diff
@@ -851,11 +851,41 @@ def render_pt(
                 top_left,
                 scale,
             )[0][0],
-            np.maximum(self.edge_thickness, 1),
+            np.maximum(int(16 / scale**0.5), 1),
             col,
-            thickness=self.thickness,
+            thickness=-1,
```
shaneahmed, line 968: why change self.thickness to a static value?

tiatoolbox/visualization/tileserver.py
```diff
@@ -513,7 +513,7 @@ def change_overlay(self: TileServer) -> str:
         overlay_path = self.decode_safe_name(overlay_path)
 
         if overlay_path.suffix in [".jpg", ".png", ".tiff", ".svs", ".ndpi", ".mrxs"]:
-            layer = f"layer{len(self.pyramids[session_id])}"
+            layer = overlay_path.stem
```
shaneahmed: Will there be a conflict if multiple overlays have same file name with a different extension?

shaneahmed (approved):
Thanks @measty It looks good to me. 

I have added a couple of comments and there are some conflicts to resolve before we merge.

tiatoolbox/tools/pyramid.py
```diff
@@ -29,6 +29,7 @@
 
 if TYPE_CHECKING:  # pragma: no cover
     from collections.abc import Iterator
+    from typing import Literal
```
shaneahmed: It's already imported on line 20.

tiatoolbox/utils/visualization.py
measty: The reason for these changes was to make points be more consistently visible across zoom levels. The thickness=-1 specifically is so that the point is filled. self.thickness value is for the thickness of boundary lines around polygons, but as a point-like object without a boundary, the value doesn't make sense to be used here.

tests/test_app_bokeh.py
measty: done

tiatoolbox/visualization/tileserver.py
measty: It would. Hopefully no one would do this, but incase they do, I've changed it so that if the stem alreadys exists in layers it will use the full name to disambiguate

PR #842, opened 2024-08-02
🧑‍💻 Fix `mypy` Type Error in `tiatoolbox/tools`

Jiaqi-Lv:
Revisited the files in `tools/`, and fixed the errors that were not detected previously.


shaneahmed (approved):
Thanks @Jiaqi-Lv

PR #846, opened 2024-08-15
Fix `mypy` type errors in `tiatoolbox/cli`

Jiaqi-Lv:


tiatoolbox/cli/common.py
```diff
@@ -397,13 +397,13 @@ def __init__(
         **kwargs: dict[str, Any],
     ) -> None:
         """Initialize TIAToolboxCLI."""
-        super().__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)  # type: ignore[arg-type]
```
Jiaqi-Lv, line 400: This is a bit tricky to fix, maybe come back to this in the future

shaneahmed (approved):
Thanks @Jiaqi-Lv

Issue #837, opened 2024-07-19
Image reader `read_region` returns read-only np.ndarray


ziyuanzhao2000:
- TIA Toolbox version: 1.5.1
- Python version: 3.9
- Operating System:

### Description

This is more like a question rather than reporting a bug. 

I wonder if there's any reason that `read_region` and similar methods seem to return a read-only array? I understand that it might be for avoiding accidental tampering with research data, but it seems inefficient to have to copy the array, for example, to use the stain normalizers (like Vahadane) on the image.

For example, this excerpt 
```
stain_normalizer = stainnorm.VahadaneNormalizer()
stain_normalizer.fit(target_image)

normed_sample = stain_normalizer.transform(sample.copy())
```
is from the stain normalizer notebook. I wonder if it's possible to load images with read-write access on the array, so we don't have to do that copy. Thanks for hearing out this suggestion!

PR #850, opened 2024-08-23
🐛 Fix Immutable Output from `WSIReader`

mostafajahanifar:
- Resolve issue #837 

The output of WSI readers (read_rect, read_region, etc) are immutable, i.e. Read-only. We assume that created Numpy arrays created from PIL images in the `background_composite` are mutable by default.
After Numpy v1.16.0, doing np.asarray(pil_image) will return an immutable Numpy array. Gladly, there is an easy fix for this, we should use `np.array()` instead of `np.asarray`.

### DIFF
--- a/tiatoolbox/utils/transforms.py
+++ b/tiatoolbox/utils/transforms.py
@@ -53,9 +53,9 @@ def background_composite(
     )
     composite.alpha_composite(image)
     if not alpha:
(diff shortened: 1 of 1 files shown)


shaneahmed (approved):
Thanks @mostafajahanifar

PR #865, opened 2024-09-20
✨ Add `PatchPredictor` Engine

shaneahmed:
-  Add `PatchPredictor` Engine based on `EngineABC`
- Add `return_probabilities` option to Params
- Removes `merge_predictions` option in `PatchPredictor` engine.
- Defines `post_process_cache_mode` which allows running the algorithm on `WSI`
- Add `infer_wsi` for WSI inference
- Removes `save_wsi_output` as this is not required after post processing.
- Removes `merge_predictions` and fixes docstring in EngineABCRunParams
- `compile_model` is now moved to EngineABC init
- Fixes bug with `_calculate_scale_factor`
- Fixes a bug in `class_dict` definition.
- `_get_zarr_array` is now a public function `get_zarr_array` in `misc`
- `patch_predictions_as_annotations` runs the loop on `patch_coords` instead of `class_probs`


measty (changes requested):
I left a couple of specific comments, but also a general comment - there seems to be a lot of duplication of code/docstrings/text in general, with PatchPredictor mostly just calling the superclass with all of its methods apart from 1. It is a ~500 line file which is almost functionally equivalent to the EngineABC class apart from infer_wsi() method. The proliferation of the amount of text you need to go through in various files to get down to where the actual things are being done/identify what is the key difference a specific subclass introduces, is something we should try to avoid, in my opinion.

tiatoolbox/models/engine/patch_classifier.py
```diff
@@ -0,0 +1,554 @@
...
+        raw_predictions: dict | Path,
+        **kwargs: Unpack[ClassifierRunParams],
+    ) -> dict | Path:
+        """Post process WSI output.
+
+        Takes the raw output from patch predictions and post-processes it to improve the
+        results e.g., using information from neighbouring patches.
+
+        """
+        _ = kwargs.get("return_probabilities")
```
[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/865/commits/8702aaf8e01ec49e3d78d884b43aa9b516f65ab2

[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/865/commits/8702aaf8e01ec49e3d78d884b43aa9b516f65ab2

tiatoolbox/models/engine/engine_abc.py
```diff
@@ -1015,20 +966,28 @@ def _run_patch_mode(
...
+        if save_dir:
+            msg = f"Output file saved at {out}."
+            logger.info(msg=msg)
+            return out
+
+        return out
+
     @staticmethod
     def _calculate_scale_factor(dataloader: DataLoader) -> float | tuple[float, float]:
         """Calculates scale factor for final output.
```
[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/865/commits/d5b6a8609decb8d961cc54f05d3e990dc4e9a8fc

shaneahmed:
> I left a couple of specific comments, but also a general comment - there seems to be a lot of duplication of code/docstrings/text in general, with PatchPredictor mostly just calling the superclass with all of its methods apart from 1. It is a ~500 line file which is almost functionally equivalent to the EngineABC class apart from infer_wsi() method. The proliferation of the amount of text you need to go through in various files to get down to where the actual things are being done/identify what is the key difference a specific subclass introduces, is something we should try to avoid, in my opinion.

Thanks @measty. I have removed duplicated code as well.

tiatoolbox/models/engine/patch_classifier.py
```diff
@@ -0,0 +1,553 @@
...
+            Output file name to save "zarr" or "db". If None, path to output is
+            returned by the engine.
+        patch_input_shape (tuple):
+            Shape of patches input to the model as tuple of height and width (HW).
+            Patches are requested at read resolution, not with respect to level 0,
+            and must be positive.
+        resolution (Resolution):
+            Resolution used for reading the image. Please see
+            :class:`WSIReader` for details.
+        return_labels (bool):
```
mostafajahanifar, line 65: repeated argument

mostafajahanifar: maybe remove the one above

tiatoolbox/models/engine/patch_predictor.py
```diff
@@ -1,31 +1,23 @@
...
-    from .io_config import ModelIOConfigABC
 
 
 class PatchPredictor(EngineABC):
     r"""Patch level predictor for digital histology images.
 
-    The models provided by tiatoolbox should give the following results:
+    The models provided by TIAToolbox should give the following results:
 
     .. list-table:: PatchPredictor performance on the Kather100K dataset [1]
```
mostafajahanifar, line 22: should we report results in this class as well as in PatchClassifier?

mostafajahanifar (commented):
Thanks Shan for doing this PR. All looks good to me. I only made a couple of minor comments. 

However, I also have a question: I don't understand the reason for needing "PatchPredictor" class to serve "PatchClassifer" as a parent class. Would be great if you can let me know why you have decided to do it like this (you most probably have discussed it before, but I'm very forgetful, I'm sorry :-s).

tiatoolbox/models/engine/patch_predictor.py
shaneahmed: We are reporting these results for PatchClassifier.

shaneahmed:
> Thanks Shan for doing this PR. All looks good to me. I only made a couple of minor comments.
> 
> However, I also have a question: I don't understand the reason for needing "PatchPredictor" class to serve "PatchClassifer" as a parent class. Would be great if you can let me know why you have decided to do it like this (you most probably have discussed it before, but I'm very forgetful, I'm sorry :-s).

Thanks @mostafajahanifar. This was a suggestion by Fayyaz to separate the two. In some cases, we do not need classification. For example, in semantic segmentor we need the raw predictions which are then converted to segmentation. In other cases, where we are building a pipeline where we need raw predictions instead of classification, this can simplify the pipeline by using just the predictor.

tiatoolbox/models/engine/patch_classifier.py
shaneahmed: https://github.com/TissueImageAnalytics/tiatoolbox/pull/865/commits/421ea775e7c7baef13caaba9ecac90e987c19f57

Jiaqi-Lv (approved):

eshasadia (approved):

tiatoolbox/models/engine/patch_predictor.py
```diff
@@ -1,31 +1,23 @@
-"""Defines Abstract Base Class for TIAToolbox Model Engines."""
+"""Defines PatchPredictor Engine."""
```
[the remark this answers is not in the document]
shaneahmed: `PatchPredictor` will be deprecated in future release as discussed. I have added warning when PatchPredictor initialised.

PR #883, opened 2024-11-22
🧑‍💻 Refine `PatchExtractor` Error Message

### LINKED ISSUE #881: Misleading error message
- TIA Toolbox version: latest
- Python version: 3.11
- Operating System: Ubuntu

### Description

The error message from `get_patch_extractor` is misleading, when `location_list` is of invalid type.

### What I Did

```
img = np.zeros((256,256,3))
coords = [[10,10]]
extractor = get_patch_extractor(
    'point',
    input_img = img,
    locations_list = coords,
    patch_size=38
)
TypeError: Please input correct image path or an ndarray image.
```
The problem is not the image, it's the locations list:
```
img = np.zeros((256,256,3))
coords = [[10,10]]
extractor = get_patch_extractor(
    'point',
    input_img = img,
    locations_list = np.array(coords),
    patch_size=38
)
```
This works fine.

Jiaqi-Lv:
- Fix #881

### DIFF
--- a/tiatoolbox/utils/misc.py
+++ b/tiatoolbox/utils/misc.py
@@ -531,7 +531,7 @@ def read_locations(
             out_table = pd.read_json(input_table)
             return __assign_unknown_class(out_table)
 
-        msg = "File type not supported."
+        msg = "File type not supported. Supported types: .npy, .csv, .json"
         raise FileNotSupportedError(msg)
 
     if isinstance(input_table, np.ndarray):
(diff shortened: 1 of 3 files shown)


tiatoolbox/utils/misc.py
```diff
@@ -540,7 +540,7 @@ def read_locations(
     if isinstance(input_table, pd.DataFrame):
         return __assign_unknown_class(input_table)
 
-    msg = "Please input correct image path or an ndarray image."
+    msg = "File type not supported."
```
shaneahmed, line 543: You need to say which file types are supported.

PR #887, opened 2024-11-29
📝 Add Example Notebook for Foundation Models

adamshephard:
- Add Example Notebook to explain how to use Foundation Models from `timm` module in TIAToolbox.


shaneahmed:
Please remember to update https://github.com/TissueImageAnalytics/tiatoolbox/blob/feature-extractor-example/examples/README.md

shaneahmed (commented):
Thanks @adamshephard . Please can you re-run the Jupyter notebook. This will update the title on the last image. Please update the feature_extraction.png with the new title. Please crop it to the image area before uploading.

shaneahmed:
> I just realised that in my code, like in this example notebook, we do not explicitly put the model into the `.eval()` mode. The same is true for the older example notebooks on which I based my code:
> 
> * `extract_deep_features()` and `extract_composition_features()` in https://github.com/TissueImageAnalytics/tiatoolbox/blob/develop/examples/full-pipelines/slide-graph.ipynb
> * `extract_deep_features()` and `extract_composition_features()` in https://github.com/TissueImageAnalytics/tiatoolbox/blob/develop/examples/inference-pipelines/slide-graph.ipynb
> 
> I got worried that I was using models in `train` mode and that I would need to recompute all the features I computed over the last couple of weeks, but then I checked the source code and realised that we rely on the `_infer_batch()` function, which runs `model.eval()` for all models that use it (line 173):
> 
> https://github.com/TissueImageAnalytics/tiatoolbox/blob/4a1940dc704d68e718c88007fb75fe1874f16708/tiatoolbox/models/architecture/vanilla.py#L148-L178
> 
> I think this behaviour is as it should be, but maybe adding a message specifying that the models are being put into `eval` mode would be beneficial.

Thanks @GeorgeBatch. As I am now finalising release `v1.6.0`, we can address this in the next release. Feel free to raise an issue or a PR for this.

PR #904, opened 2025-01-16
:arrow_up: Upgrade Zarr to v3+ Compatibility

measty:
## ✅ Summary

- Enables compatibility with Zarr v3+
- Simplifies error handling across Zarr versions
- Removes deprecated storage backends and APIs
- Lays groundwork for full Zarr v3 migration

This PR updates TIAToolbox to support **Zarr v3+**, addressing breaking changes introduced in Zarr ≥ 3.0 and aligning internal APIs, storage handling, and error handling accordingly.

---

Zarr v3 removed `zarr.errors.FSPathExistNotDir`. In Zarr ≥ 3.0, equivalent failure modes now raise standard Python exceptions (e.g. `FileNotFoundError`).

To avoid fragile version-specific logic:
- Removed checks for specific Zarr exception types
- Replaced them with generic exception handling that works consistently across Zarr versions

---

### Dependency updates
- 🆙 Upgrade to:
  - `zarr >= 3.0.8`
  - `tifffile >= 2025.5.21`

### Zarr API & storage migration
- Refactor `zarr.core.Array` → `zarr.Array`
- Replace `zarr.DirectoryStore` with `zarr.storage.LocalStore`
- Remove `ngff.sqlitestore` (Zarr v3 no longer supports `zarr.SQLiteStore`)
- Replace `canvas_zarr.store.path` with `canvas_zarr.store.root`
- Ensure Zarr paths explicitly use string keys (`key / str(i)` instead of `key / i`)

### Dask & array handling
- Remove unsupported `object_codec` usage in `dask.to_zarr`
- Update `zarr.resize` calls to use tuple inputs
- Replace `output.items()` with `output.members()`

---

## ⚠️ Known limitations & follow‑ups

Zarr v3 removes or significantly changes several APIs still used by TIAToolbox, including:
- `zarr.LRUStoreCache`
- `zarr.DirectoryStore`
- `zarr.SQLiteStore`

While this PR resolves immediate incompatibilities and enables Zarr v3 usage, **additional refactoring will be required** to fully replace or redesign components that depend on these removed features.

In addition, **full compatibility with `tifffile` depends on upstream Zarr v3 support**, and some workflows may remain limited until that stabilises.

📖 For a detailed overview of breaking changes, see the official Zarr v3 migration guide:  
https://zarr.readthedocs.io/en/latest/user-guide/v3_migration.html

---


shaneahmed:
Zarr v3 is not compatible with tifffile https://github.com/cgohlke/tifffile/issues/282

shaneahmed:
Zarr 3 is supported by tifffile. However, zarr 3 only supports Python 3.11+.

shaneahmed:
@aacic Current implementation of `FsspecJsonWSIReader` is not compatible with Zarr V3. I have tested multiple options but they fail. Do you have any suggestions?

https://github.com/TissueImageAnalytics/tiatoolbox/blob/750ca026df82839e1e286b2e25d17ef53aa55a85/tiatoolbox/wsicore/wsireader.py#L4392-L4459

aacic:
> @aacic Current implementation of `FsspecJsonWSIReader` is not compatible with Zarr V3. I have tested multiple options, but they fail. Do you have any suggestions?
> 
> https://github.com/TissueImageAnalytics/tiatoolbox/blob/750ca026df82839e1e286b2e25d17ef53aa55a85/tiatoolbox/wsicore/wsireader.py#L4392-L4459

Please give me some time to look into this. I'll try to look at this next week. I don't know it from the top of my head.

shaneahmed:
All the errors have been fixed except `FsspecJsonWSIReader`.

shaneahmed:
> All the errors have been fixed except `FsspecJsonWSIReader`.

The updates to engines output breaks, `test_hovernet_on_box` which also needs to be fixed.

aacic:
> All the errors have been fixed except `FsspecJsonWSIReader`.

@shaneahmed  I was able to make it work locally. I need some more time to verify the solution, and I'll try to submit a PR sometime next week. I'll keep you posted.

shaneahmed:
> > All the errors have been fixed except `FsspecJsonWSIReader`.
> 
> @shaneahmed I was able to make it work locally. I need some more time to verify the solution, and I'll try to submit a PR sometime next week. I'll keep you posted.

Thank you @aacic That would be wonderful.

shaneahmed:
> > All the errors have been fixed except `FsspecJsonWSIReader`.
> 
> The updates to engines output breaks, `test_hovernet_on_box` which also needs to be fixed.

@measty Please would you be able to check the error with ` test_hovernet_on_box`? This is probably due to updates in the way new zarr output is saved.

JiwaniZakir (commented):
The padding-removal logic added in `assert_annotation_store_patch_output` and `assert_qupath_json_patch_output` uses `np.iinfo(contours.dtype).min` as the sentinel value, but `np.iinfo` only accepts integer dtypes — if `contours.dtype` is ever a float (e.g., `float32` or `float64`), this will raise a `ValueError` at runtime. A safer approach would be to guard with `np.iinfo` only when `np.issubdtype(contours.dtype, np.integer)` and fall back to `np.finfo` or a different sentinel otherwise.

The pervasive switch from integer to string keys (e.g., `output_zarr["x"][0]` → `output_zarr["x"]["0"]` in `test_nucleus_detection_engine.py` and the `str(i)` conversion in `assert_output_lengths`) reflects a real zarr v3 API change, but it's worth verifying this pattern is consistently applied in the production code paths as well — the tests only cover the assertion helpers, and any production code that still uses integer indexing into zarr groups would silently fail or return wrong results.

The switch from `output_.items()` to `output_.members()` in `test_semantic_segmentor.py` is correct for zarr v3, but the truncated diff makes it unclear whether similar `items()` calls in production code (not just tests) have been updated — a quick grep for `.items()` on zarr group objects in the main source would be worth confirming.

tests/engines/test_multi_task_segmentor.py
```diff
@@ -1205,7 +1209,10 @@ def assert_output_lengths(
     """Assert lengths of output dict fields against expected counts."""
     for field in fields:
         for i, expected in enumerate(expected_counts):
-            assert len(output[field][i]) == expected, f"{field}[{i}] mismatch"
+            idx = str(i) if isinstance(output[field], (zarr.Array, zarr.Group)) else i
```
shaneahmed:
```suggestion
            idx = str(i) if isinstance(output[field], zarr.Group) else i
```

tests/engines/test_multi_task_segmentor.py
```diff
@@ -1263,11 +1270,21 @@ def assert_output_equal(
     """Assert equality of arrays across outputs for given fields/indices."""
     for field in fields:
         for i_a, i_b in zip(indices_a, indices_b, strict=False):
-            left = output_a[field][i_a]
-            right = output_b[field][i_b]
+            i_a_ = (
+                str(i_a)
+                if isinstance(output_a[field], (zarr.Array, zarr.Group))
```
shaneahmed:
```suggestion
                if isinstance(output_a[field], zarr.Group)
```

tests/engines/test_multi_task_segmentor.py
```diff
@@ -1263,11 +1270,21 @@ def assert_output_equal(
...
+                else i_a
+            )
+            i_b_ = (
+                str(i_b)
+                if isinstance(output_b[field], (zarr.Array, zarr.Group))
```
shaneahmed:
```suggestion
                if isinstance(output_b[field], zarr.Group)
```

shaneahmed (approved):
I have tested the engines locally and they work well. I have also addressed Co-Pilots comments.

tests/engines/test_multi_task_segmentor.py
```diff
@@ -1534,3 +1564,248 @@ def test_cli_model_single_file(remote_sample: Callable, track_tmp_path: Path) ->
...
+def patch_save_qupath_json(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Patch save_qupath_json so compute_qupath_json returns JSON directly."""
+
+    def fake_save_qupath_json(
+        save_path: Path | None,  # noqa: ARG001
+        qupath_json: dict[str, Any],
+    ) -> dict[str, Any]:
+        return qupath_json
+
+    import tiatoolbox.models.engine.multi_task_segmentor as mts  # noqa: PLC0415
```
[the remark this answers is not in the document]
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/904/commits/ccb0a14dcfd144d26cc210c0c54f42bfcab6f6e5

PR #905, opened 2025-01-16
📌 Pin `zarr<3.0.0`

measty:
Zarr 3.0 breaks lots of things, probably wont be a quick fix to make toolbox compatible with it, so probably best to pin <3.0 for now

See: https://zarr.readthedocs.io/en/latest/user-guide/v3_migration.html  and #904 for more details


shaneahmed (approved):
Thanks @measty

PR #911, opened 2025-02-14
🐛 Fix in `test_arch_mapde` and `test_arch_sccnn`

Jiaqi-Lv:
- If cuda is available model should be moved to cuda otherwise tests will fail as test data is moved to cuda.


shaneahmed (approved):
Thanks @Jiaqi-Lv

PR #914, opened 2025-02-21
:bug: Fix `MapDe` `dist_filter` Shape 

Jiaqi-Lv:
- Fix `dist_filter` in `MapDe` model for multi-class output.

Explanation:
Previously, if we set `num_class` to more than 1, the model would still output 1 channel. This was because the `dist_filter` always had size of 1 in its first dimension, however the first dimension determines the number of output channels in the tensor produced by `torch.functional.F.conv2d`.
I changed this by simply repeating the filters the match the number of output classes.


Jiaqi-Lv:
Not sure why codecov failed

shaneahmed (approved):
Thanks @Jiaqi-Lv

PR #923, opened 2025-04-16
🐛 Fix Multi-GPU Support with `torch.compile`

adamshephard:
Change multi-GPU mode from `DataParallel` to `DataDistributedParallel` to work with torch.compile. However, this essentially limits the task to using one GPU alone when using torch.compile. It is not a trivial solution to change this to use multiple GPUs that also work with torch.compile. We will release a future fix to fully correct this with the new engine.


shaneahmed (approved):
Thanks @adamshephard

PR #926, opened 2025-04-25
:sparkles: Add `dict_to_store_semantic_segmentor`

shaneahmed:
- Add a function to convert the output of TIAToolbox segmentation engine to AnnotationStore.


shaneahmed:
@adamshephard @measty @mbasheer04 I have created this PR separately so we can work towards different scenarios particularly  where the annotations have holes.

adamshephard:
This still doesn't handle the holes issue. In the last test I have added, you can see that the inner contour/hole is treated the same as the outer contour. When these contours are drawn, any holes present will be filled in as a result by the hole contour.

PR #866, opened 2024-09-20
:sparkles: Define `SemanticSegmentor` with the New `EngineABC`

shaneahmed:
- Define `SemanticSegmentor` with the New `EngineABC`

PR #825, opened 2024-06-21
ENH: initial draft multichannel reading

measty:
Initial draft implementation of multi-channel reading.

multi_channel

Still needs lots of work so its a draft at the moment.

PR #929, opened 2025-04-30
✨ Add Feature to Write Model Probabilities as `ome.tiff` for Overlay

shaneahmed:
- This PR saves a NumPy or Zarr Array to OME tiff. This will help to save probability maps from deep learning models as ome.tiff which can be visualized using TIAViz.
- The image is written as tiles and stitched in OME tiff along with associated necessary metadata for heatmaps generated by segmentation engine.  #866 
- Reading with QuPath has been tested.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import hashlib
 import shutil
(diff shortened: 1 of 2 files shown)


shaneahmed:
The output file should be readable using multichannel viewer in #825

PR #867, opened 2024-09-20
:art: Refactor `ModelABC` to Help Use Default Torch Models

shaneahmed:
- Refactor `ModelABC` to Help Use Default Torch Models
- Remove `infer_batch` from `ModelABC`

PR #930, opened 2025-05-01
`mypy` type check `vanilla.py`

Jiaqi-Lv:
Updated type annotations for `vanilla.py`


Jiaqi-Lv:
Maybe I should come back to this after `ModelABC` is refactored  #867

Issue #933, opened 2025-05-05
Unable to open entire DICOM WSI series


fedorov:
I have a folder that contains DICOM instances corresponding to the different levels of the pyramid as well as overview/label images. Is it the case that loading of the entire pyramid is not possible? If I load a single instance, I see only one level in the loaded object, and loading of the folder fails with an error.

Here's code to reproduce:

```
# download a DICOM WSI series from Imaging Data Commons
!pip install --upgrade idc-index
!mkdir idc_sample && idc download --download-dir ./idc_sample 1.3.6.1.4.1.5962.99.1.1276379089.2049488480.1714968362961.4.0

reader = WSIReader.open("/content/idc_sample/cmb_aml/MSB-01723/2.25.93893694229316003747676080569358410213/SM_1.3.6.1.4.1.5962.99.1.1276379089.2049488480.1714968362961.4.0/113fed92-2815-4d5d-80f6-2b6555cbb51d.dcm")

info_dict = reader.info.as_dict()
pprint(info_dict)
```

Error:

```
|2025-05-05|19:23:40.535| [WARNING] Raw data is None.
{'axes': 'YXS',
 'file_path': None,
 'level_count': 1,
 'level_dimensions': ((1991, 2048),),
 'level_downsamples': [1.0],
 'mpp': (4.038700803616, 4.038700803616),
 'objective_power': None,
 'slide_dimensions': (1991, 2048),
 'vendor': 'Leica Biosystems'}
```

We have over 35TB of publicly available slide microscopy images in IDC, all in DICOM format https://testing-portal.canceridc.dev/explore/filters/?Modality_op=OR&Modality=SM, and it would be quite helpful for our users to be able to use `tiatoolbox` with this data directly.

fedorov:
The aforementined PR added `.dcm` to the list of recognizable files. I installed tiatoolbox from current hash, and trying to open a DICOM WSI pyramid results in the following error. Should I open a separate issue about this?

You can access the sample that I was using in testing as follows:
```bash
$ pip install --upgrade idc-index
$ idc download --dir-template="" 2.25.37411452242028229091597727833436758051
```

```
|2025-06-11|12:30:09.339| [INFO] WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
 * Running on http://127.0.0.1:5000
|2025-06-11|12:30:09.340| [INFO] Press CTRL+C to quit
[...]

PR #934, opened 2025-05-08
🐛 Fix DICOM Reading (Issue #933)

measty:
In response to issue #933, I tried the dicom image they linked. The reason they couldnt read the pyramid was simply because our DICOMWSIReader expects the folder that contains the levels to be passed as the image path, not the path to an idividual file within the dicom folder. So, their image reads within tiatoolbox just by passing the folder path.

However, while investigating this I found a bug with tiatoolbox dicom reader in that it is not reading at the correct location (wsidicom readregion expects the location in the coordinate space of the resolution being read, and tiatoolbox reader was giving it location in baseline)

This PR fixes that bug (and adds folders that end in .dcm as a valid slide extension for TIAViz so dicom can be viewed in there)

Update -  Also updates the parameterized WSIReader tests as follows, to make them more robust so hopefully this sort of bug doesnt slip through again.

All the 'test read consistency' type checks that read regions of the image in different ways that should be equivalent if the reader is working as intended, no longer just read a tile located at (0, 0) which: 
1) may not even have any tissue in (making checks like structural similarity meaningless)
2) may not expose bugs related to scaling of location as (0, 0) is a special case in this respect

Instead, these checks now find the center-of-mass of the tissue mask of the WSI, and use a tile located around that instead, to ensure a meaningful tile not located at (0,0) is used. With these changes, the current develop branch will fail the tests, but this bug-fixed branch will pass.

### DIFF
--- a/tiatoolbox/visualization/bokeh_app/main.py
+++ b/tiatoolbox/visualization/bokeh_app/main.py
@@ -725,7 +725,16 @@ def populate_slide_list(slide_folder: Path, search_txt: str | None = None) -> No
     """Populate the slide list with the available slides."""
     file_list = []
     len_slidepath = len(slide_folder.parts)
-    for ext in ["*.svs", "*ndpi", "*.tiff", "*.mrxs", "*.jpg", "*.png", "*.tif"]:
+    for ext in [
+        "*.svs",
+        "*ndpi",
+        "*.tiff",
+        "*.mrxs",
+        "*.jpg",
+        "*.png",
+        "*.tif",
+        "*.dcm",
+    ]:
         file_list.extend(list(Path(slide_folder).glob(str(Path("*") / ext))))
         file_list.extend(list(Path(slide_folder).glob(ext)))
     if search_txt is None:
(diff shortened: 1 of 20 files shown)


fedorov:
> The reason they couldnt read the pyramid was simply because our DICOMWSIReader expects the folder that contains the levels to be passed as the image path, not the path to an idividual file within the dicom folder. So, their image reads within tiatoolbox just by passing the folder path.

Thank you for working on this! I thought I tried opening it as a folder, and I thought it didn't work, but I confirm it is working for me as you described. Probably I missed something in my initial testing.

shaneahmed (approved):
Thanks @measty Looks good.

tiatoolbox/visualization/bokeh_app/main.py
```diff
@@ -2086,7 +2095,16 @@ def setup_doc(self: DocConfig, base_doc: Document) -> tuple[Row, Tabs]:
 
         # Set initial slide to first one in base folder
         slide_list = []
-        for ext in ["*.svs", "*ndpi", "*.tiff", "*.tif", "*.mrxs", "*.png", "*.jpg"]:
+        for ext in [
+            "*.svs",
+            "*ndpi",
```
fedorov, line 2100: should this be `".ndpi"` - is the dot missing?

PR #937, opened 2025-05-23
:pushpin: Pin `tifffile<=2025.5.10`

Jiaqi-Lv:
This PR refines the dependency version requirements by pinning tifffile with an explicit upper bound.  
- Pin tifffile with both a minimum version (2022.10.10) and maximum version (2025.5.10).  
- This ensures controlled compatibility for tifffile across environments.


shaneahmed (approved):
Thanks @Jiaqi-Lv

PR #947, opened 2025-08-08
make patchdataset picklable

measty:
Delays the creation of reader object till first call to __get_item__ so that it remains picklable till then.

Allows n_loader_workers > 0 to be used on windows without an error.

Speedup: 2min 48 sec with 0 loader workers for patch_prediction example notebook -> 1min 13 sec with 4 workers


shaneahmed (approved):
Thanks @measty This looks good.

PR #964, opened 2025-10-23
:new: Define `NucleusInstanceSegmentor`

shaneahmed:
- Define `NucleusInstanceSegmentor`


shaneahmed:
Nucleus Instance segmentor is implemented in #981

PR #968, opened 2025-11-07
🆕 Simplified `SAM` Segmentation in `TIAViz`

measty:
This PR takes the sam-viz PR and fixes and substantially simplifies it.

I think the original PR was going for a more general prompt segmentor engine that would in theory cope with prompts dispersed over a whole WSI, but I think that approach has issues in practice (and it would also need some significant work to fix from its current state, I think). This PR focusses on segmenting prompts made within a specific view window, which I think is more practical and allows for a much simplified approach.

To use, draw a few points (and/or a box) on some object(s) in the field of view in TIAViz using this widget:
image
Then select SAM in the model select box and click run. The segmetations should be displayed after a short time.

Examples:
image
image
image


tiatoolbox/models/engine/prompt_segmentor.py
```diff
@@ -0,0 +1,115 @@
...
+        model: torch.nn.Module = None,
+    ) -> None:
+        """Initializes the PromptSegmentor."""
+        if model is None:
+            model = SAM()
+        self.model = model
+        self.scale = 1.0
+        self.offset = (0, 0)
+
+    def predict(  # skipcq: PYL-W0221
```
shaneahmed:
```suggestion
    def run(  # skipcq: PYL-W0221
```

tiatoolbox/models/engine/prompt_segmentor.py
```diff
@@ -0,0 +1,115 @@
...
+        self,
+        imgs: list,
```
shaneahmed:
```suggestion
        images: list,
```

measty:
PR is now ready for review

.pre-commit-config.yaml
```diff
@@ -2,6 +2,8 @@ ci:
   autofix_prs: true
   autoupdate_commit_msg: ':technologist: pre-commit autoupdate'
   autoupdate_schedule: 'monthly'
+default_language_version:
+  python: python3.12
```
shaneahmed:
```suggestion
  python: python3.10
```

Can we set this to the base supported version i.e., 3.10.

tiatoolbox/visualization/bokeh_app/main.py
```diff
@@ -66,8 +66,9 @@
 # GitHub actions seems unable to find TIAToolbox unless this is here
 sys.path.insert(0, str(Path(__file__).parent.parent.parent.parent))
 from tiatoolbox import logger
-from tiatoolbox.models.engine.nucleus_instance_segmentor import (
-    NucleusInstanceSegmentor,
+from tiatoolbox.models.engine.nucleus_instance_segmentor import NucleusInstanceSegmentor
+from tiatoolbox.models.engine.prompt_segmentor import (
+    PromptSegmentor,  # skipcq: FLK-E402
```
shaneahmed: It's strange that you are getting `FLK-E402` error on this.

tiatoolbox/visualization/bokeh_app/main.py
```diff
@@ -1276,6 +1279,104 @@ def segment_on_box() -> None:
...
+        )
+        / scale_factor
+        if len(x) > 0
+        else None
+    )
+
+    # Get box coordinates
+    x = np.round(UI["box_source"].data["x"])
+    y = np.round(UI["box_source"].data["y"])
+    height = np.round(UI["box_source"].data["height"])
```
shaneahmed: height and width is being overwritten on 1329 and 1330

tests/models/test_prompt_segmentor.py
```diff
@@ -0,0 +1,53 @@
+"""Unit test package for prompt segmentor."""
```
shaneahmed: Please move this file to `tests/engines`

tiatoolbox/models/architecture/sam.py
```diff
@@ -0,0 +1,238 @@
...
+
+        return image_masks, image_scores
+
+    def forward(  # skipcq: PYL-W0221
+        self: SAM,
+        imgs: list,
+        point_coords: list | None = None,
+        box_coords: list | None = None,
+    ) -> tuple[np.ndarray, np.ndarray]:
+        """Torch method. Defines forward pass on each image in the batch.
```
shaneahmed:
```suggestion
        """PyTorch method. Defines forward pass on each image in the batch.
```

tiatoolbox/models/engine/prompt_segmentor.py
```diff
@@ -0,0 +1,114 @@
...
+    one of either point_coords or box_coords must be provided to guide
+    segmentation.
+
+    Args:
+        model (SAM):
+            Model architecture to use. If None, defaults to SAM.
+
+    """
+
+    def __init__(
```
shaneahmed, line 35: Do we need any of these inputs to match with the new EngineABC signature?

    def __init__(
        self: EngineABC,
        model: str | ModelABC,
        batch_size: int = 8,
        num_workers: int = 0,
        weights: str | Path | None = None,
        *,
        device: str = "cpu",
        verbose: bool = False,

tiatoolbox/models/engine/prompt_segmentor.py
```diff
@@ -0,0 +1,114 @@
...
+        self.scale = 1.0
+        self.offset = (0, 0)
+
+    def run(  # skipcq: PYL-W0221
+        self,
+        images: list,
+        point_coords: np.ndarray | None = None,
+        box_coords: np.ndarray | None = None,
+        save_dir: str | Path | None = None,
+        device: str = "cpu",
```
shaneahmed, line 51: Is there an output type option if a user wants to save the results?

tiatoolbox/models/engine/prompt_segmentor.py
```diff
@@ -0,0 +1,114 @@
...
+                image patch.
+            save_dir (str or Path):
+                Directory to save the output databases.
+            device (str):
+                Device to run inference on.
+
+        Returns:
+            list[Path]:
+                Paths to the saved output databases.
+        """
```
shaneahmed, line 72:
```suggestion

        """
```

shaneahmed (changes requested):
Thanks @measty 

This looks good. A few minor comments if you could address before we merge. 

- I was not able to test it though on my workstation using `--base-path` argument. 
- Also it would be good to show any progress bar or message on the visualization rather than only in the terminal if possible.
- Can we reset the model weights after the first run if possible?

tests/models/test_prompt_segmentor.py
measty: done

tiatoolbox/models/engine/prompt_segmentor.py
measty: It is just saved as annotation store. As it is primarily designed for interactive use in TIAViz, and annotation store is the most sensible output for segmentations anyway, i think that is suffiecient at least for this initial implementation.

tiatoolbox/models/engine/prompt_segmentor.py
measty: I don't think so, as this doesn't inherit from EngineABC. The prompt based segmentation on regions is sufficiently different from how all the other engines operate that it wouldn't make much sense to inherit from that.

adamshephard (approved):

shaneahmed (approved):
Thanks @measty

PR #989, opened 2026-01-28
:bug: Fix Division by Count in `SemanticSegmentor`

shaneahmed:
## 🐛 Fix Division by Count in `SemanticSegmentor`

This PR fixes a core issue in patch merging that caused incorrect normalization of segmentation outputs, and includes several related improvements for consistency and correctness.

### 1. Correct patch‑merge count accumulation
- The `count` array was indexed incorrectly, so only the first patch row was normalized properly.
- Updated indexing ensures accurate per‑pixel accumulation across all rows.

### 2. Add `class_dict` to all models
- Introduces `class_dict` in `ModelABC`.
- Ensures `SemanticSegmentor` can reliably use class dictionaries when none are passed explicitly.

### 3. Improve output type handling
- `SemanticSegmentor.run()` now correctly indicates that it may return a `list[Path]` when multiple `.db` files are produced.

### 4. More flexible probability storage
- `store_probabilities()` now accepts a `name` parameter, enabling multiple probability datasets in Zarr outputs.

### 5. Test updates
- Adjusted expected nucleus counts and probability means to reflect corrected normalization.
- Updated CLI tests to use remote samples and ensure proper cleanup.

### DIFF
--- a/tests/engines/test_nucleus_detection_engine.py
+++ b/tests/engines/test_nucleus_detection_engine.py
@@ -264,7 +264,7 @@ def test_nucleus_detector_wsi(remote_sample: Callable, track_tmp_path: Path) -> 
     )
 
     store = SQLiteStore.open(save_dir / "wsi4_512_512.db")
-    assert 255 <= len(store.values()) <= 265
+    assert 245 <= len(store.values()) <= 255
     annotation = next(iter(store.values()))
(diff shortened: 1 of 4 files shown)


Jiaqi-Lv (approved):
Looks Good.

PR #1006, opened 2026-02-25
:green_heart: Improve Performance Using Mask

shaneahmed:
## Summary
This PR updates the WSI segmentation workflow so that whole‑slide images are processed **only within the tissue‑mask bounds**, and the resulting predictions are **padded back to full-slide dimensions** once inference completes. It also adds support for supplying a global **offset** to all model post‑processing steps, ensuring instance and semantic outputs (centroids, contours, boxes) remain aligned with the original slide coordinates. The dataset/engine API is made more consistent (`mask_path` → `input_mask`), and tests are expanded to cover masked, non-square, and tile-based scenarios.

### Mask-bounded processing and output padding
- Engines compute `mask_bounds`, filter output locations to those inside the mask, and track required padding.
- Final probabilities and predictions are padded using Dask to restore the original full-slide output shape.
- Supports both semantic and multi-task segmentation paths.

### Offset-aware post-processing
- `HoVerNet`, `HoVerNetPlus`, and `MicroNet` now accept an `offset` argument and apply it to all geometric outputs.
- Engines automatically provide the appropriate offset based on mask origin or tile layout.

### API and docstring updates
- `mask_path` is renamed to `input_mask` and now accepts either a file path or an `np.ndarray` mask.
- Docstrings clarify default behaviors for `auto_get_mask`, probabilities, memory thresholds, and worker counts.
- Type hints and parameter descriptions improved for consistency across dataset and engine classes.

### Tile merging and instance ID handling
- Horizontal merge functions support a fixed `canvas_width` to avoid allocating overly large row canvases.
- Tile-based instance merging now preserves unique instance IDs across tiles and handles overlaps more safely.

### Test updates
- Tests updated to use `input_mask`, verify masked execution paths, validate shapes and probability behavior, and cover tile mode and offset-aware post-processing.


Jiaqi-Lv (approved):

tiatoolbox/models/engines/semantic_segmentor.py
```diff
@@ -1629,3 +1745,71 @@ def prepare_full_batch(
...
+    pad_left: int,
+    pad_right: int,
+) -> tuple[tuple, ...]:
+    """Build pad width for dask padding."""
+    pad_width = []
+    for axis in range(arr.ndim):
+        if axis == 0:  # height
+            pad_width.append((pad_top, pad_bottom))
+        elif axis == 1:  # width
+            pad_width.append((pad_left, pad_right))
```
measty: calling this pad_width when width/height are being used simultanously in a different way is potentially consfusing.

its looking at padding in width, in height and also maybe channel, so maybe calling the function build_pad_size, and then pad_size = [] etc would be better

tiatoolbox/models/engines/semantic_segmentor.py
```diff
@@ -608,6 +630,85 @@ def infer_wsi(
...
+        Args:
+            raw_predictions (dict[str, da.Array]):
+                Dictionary containing raw model predictions as Dask arrays.
+            save_path (Path):
+                Path to save the intermediate output. The intermediate output is saved
+                in a zarr file.
+            **kwargs (PredictorRunParams):
+                Additional runtime parameters to configure prediction.
+
+                Optional Keys:
```
measty, line 653: These optional keys docstrings (from RunParams classes) are getting duplicated around a lot, and then sometimes being edited in one place but not in another place. It would be better (maybe a later PR) to have the docstring for each RunParams-style class exist once somewhere, and then wherever the kwargs need to be included in a docstring you just grab them from there. That avoids inconsistencies sneaking in where the kwargs docstring gets modified in one place but not another

measty (commented):
there is some issue with the instance merging:

Using this PR:
Image

Older run from aweek or so ago (multitask-segmentor branch):

Image

Somehow this PR is resulting in multiple overlapping cell fragments in places - those ones near the bottom with the band across them are actually 3 separate contours

tiatoolbox/models/engines/semantic_segmentor.py
shaneahmed: pad_width is a dask argument name. I could rename it to dask_pad_width.

tiatoolbox/models/engines/semantic_segmentor.py
shaneahmed: We can limit docstrings to only relevant kwargs. But this can be sorted in a later PR.

tiatoolbox/models/engines/semantic_segmentor.py
shaneahmed: Addressed in this commit https://github.com/TissueImageAnalytics/tiatoolbox/pull/1006/commits/631ab74003fc2a57ae4026c2430d5d436891482e

Renamed `pad_width` to `da_pad_width`.

Issue #1068, opened 2026-05-24
Writing to annotationstore implicitly drops class 0


mraoaakash:
Running a prediction using output type as annotationstore removes class '0' implicitly.

- **Model used:** fcn_resnet50_unet-bcss
- **Ioconfig used:**
  - "input_resolutions": [{"units": "mpp", "resolution": 0.25}], 
  - "output_resolutions": [{"units": "mpp", "resolution": 0.25}],
  - "patch_input_shape": (1024, 1024),
  - "patch_output_shape": (512, 512),
  - "stride_shape": (256, 256),
  - "save_resolution": {"units": "mpp", "resolution": 0.25}, 

Passing zarr to the output_type yields the following label proportions:
`{0: 14736243, 1: 91068618, 2: 5553089, 3: 1884865, 4: 31333761}`

However, using annotationstore yields 
`{'Inflamatory': 2672, 'Necrosis': 1955, 'Others': 5767, 'Stroma': 5065} `

The class_mapping for fcn_resnet50_unet-bcss is: 
```
class_mapping:
      Tumour: 0
      Stroma: 1
      Inflamatory: 2
      Necrosis: 3
      Others: 4
```
Class 0, tumour is dropped from the annotation store but exists in the zarr output.

Tests to verify:
1. Passed explicit class_mapping to the run function - **fails** 
2. Passed default ioconfig from model yaml file - **fails**
3. Passed no ioconfig - **fails**
4. Pass zarr instead of annotationstore - **passes**

mraoaakash:
Bypassing current store without use of ignore_index, bypassing entirely by using a function that skips the ignore index tests gives valid predictions:

### Function:
```
def convert_semantic_zarr_to_annotation_store(
    zarr_path: str | Path,
    annotation_path: str | Path,
    class_dict: dict[int, str] | None = None,
    scale_factor: tuple[float, float] = (1.0, 1.0),
    offset: np.ndarray | None = None,
    verbose: bool = True,
) -> Path:
    """Convert TIAToolbox semantic segmentation zarr predictions to AnnotationStore."""
    import zarr
    from tiatoolbox.utils.misc import dict_to_store_semantic_segmentor

    output_path = Path(annotation_path)
    ensure_dir(str(output_path))
    converted_path = dict_to_store_semantic_segmentor(
        patch_output=zarr.open(str(zarr_path), mode="r"),
        scale_factor=scale_factor,
        output_type="annotationstore",
[...]

PR #1069, opened 2026-05-24
:bug: Remove `ignore_index` for `fcn_resnet50_unet-bcss`

mraoaakash:
In fcn_resnet50_unet-bcss ignore_index is defined as 0 which corresponds to the tumor class which leads to exclusion of tumor class during annotation store writing. 

Addresses #1068


Jiaqi-Lv (approved):
Looks good. I've also updated the `class_dict` for `fcn_resnet50_unet-bcss` and `efficientunet-tissue-mask` models so they can get automatically picked up by the engine.

mraoaakash:
I think #1068 can be extended to have an enhancement as well after this (added a tag). It might be worth extending this to assign `class_dict`  or equivalent to the 
- `kather100k` models
- `pcam` models
- `unet_tissue_mask_tsef`
- `hovernet_original-kumar`
- `micronet-consep`
- `nuclick` models

Or whichever of these is relevant.

Jiaqi-Lv:
> I think #1068 can be extended to have an enhancement as well after this (added a tag). It might be worth extending this to assign `class_dict` or equivalent to the
> 
> * `kather100k` models
> * `pcam` models
> * `unet_tissue_mask_tsef`
> * `hovernet_original-kumar`
> * `micronet-consep`
> * `nuclick` models
> 
> Or whichever of these is relevant.

I think we should definitely do this for all the models. Someone will need to go through every model and figure out the output classes and may need to update some of the unit tests. Would you like to volunteer @mraoaakash :laughing: ?

mraoaakash:
> > I think #1068 can be extended to have an enhancement as well after this (added a tag). It might be worth extending this to assign `class_dict` or equivalent to the
> > 
> > * `kather100k` models
> > * `pcam` models
> > * `unet_tissue_mask_tsef`
> > * `hovernet_original-kumar`
> > * `micronet-consep`
> > * `nuclick` models
> > 
> > Or whichever of these is relevant.
> 
> I think we should definitely do this for all the models. Someone will need to go through every model and figure out the output classes and may need to update some of the unit tests. Would you like to volunteer @mraoaakash 😆 ?

Yes 👀😆

PR #1074, opened 2026-06-04
🐛 Fix multi-gpu error (AttributeError) for `kongnet` and `nucleus_detector`

gozdeg:
Follow-up to the earlier multi-GPU fix. kongnet and nucleus_detector were missed and still crash on multi-GPU machines with:

AttributeError: 'DataParallel' object has no attribute 'target_channels'

When multi-gpu, the model is wrapped in `nn.DataParallel` / `DistributedDataParallel`, so these need to go through **the `_get_model_attr()` wrapper helper function** (or a similar approach) which unwraps the module before reading the attribute.

### Changes
- `nucleus_detector.py`: read `min_distance` / `tile_shape` via the `_get_model_attr` helper instead of directly off `self.model`.
- `kongnet.py`: unwrap the module before reading `target_channels` in `infer_batch`. This can't reach the helper function, so it applies the same unwrap inline.
- `engine_abc.py`: simplify _get_model_attr to a single getattr-based unwrap and clarify its docstring.

### DIFF
--- a/tiatoolbox/models/architecture/kongnet.py
+++ b/tiatoolbox/models/architecture/kongnet.py
@@ -862,10 +862,8 @@ class KongNet(ModelABC):
         imgs = imgs.to(device=device, dtype=torch.float32)
         imgs = imgs.permute(0, 3, 1, 2)  # to NCHW
 
-        try:
-            target_channels = model.target_channels
-        except AttributeError:
-            target_channels = model.module.target_channels
+        # unwrap DataParallel/DDP if present (happens in multi-gpu settings)
+        target_channels = getattr(model, "module", model).target_channels
 
         with torch.inference_mode():
             logits = model(imgs)
--- a/tiatoolbox/models/architecture/kongnet.py
+++ b/tiatoolbox/models/architecture/kongnet.py
@@ -862,9 +862,14 @@ class KongNet(ModelABC):
         imgs = imgs.to(device=device, dtype=torch.float32)
         imgs = imgs.permute(0, 3, 1, 2)  # to NCHW
 
+        try:
+            target_channels = model.target_channels
+        except AttributeError:
+            target_channels = model.module.target_channels
+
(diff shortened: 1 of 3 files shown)


tiatoolbox/models/architecture/kongnet.py
```diff
@@ -862,9 +862,14 @@ def infer_batch(
...
 
+        try:
+            target_channels = model.target_channels
+        except AttributeError:
+            target_channels = model.module.target_channels
+
         with torch.inference_mode():
             logits = model(imgs)
-            target_logits = logits[:, model.target_channels, :, :]
+            target_logits = logits[:, target_channels, :, :]
```
shaneahmed, line 870: Why not `_get_model_attr`?

gozdeg: @shaneahmed _get_model_attr() was defined in engineABC, I don't think we can reach it from here

shaneahmed: Can you try this?

```python
if hasattr(model, "target_channels"):
  target_channels = model.target_channels
elif hasattr(model.module, "target_channels"):
  target_channels = model.module.target_channels
else:
  raise AttributeError:
```

tiatoolbox/models/architecture/kongnet.py
```diff
@@ -862,9 +862,14 @@ def infer_batch(
         imgs = imgs.to(device=device, dtype=torch.float32)
         imgs = imgs.permute(0, 3, 1, 2)  # to NCHW
 
+        try:
+            target_channels = model.target_channels
+        except AttributeError:
+            target_channels = model.module.target_channels
+
```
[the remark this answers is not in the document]
gozdeg: something similar is done in last commit

shaneahmed (approved):
Thanks @gozdeg

PR #1075, opened 2026-06-05
♻️ Refactor / Split `wsireader` into Multiple Modules

shaneahmed:
- Splits wsireader into multiple modules.


shaneahmed:
This PR will be split into multiple PRs, see #1087 , #1088

PR #1089, opened 2026-07-02
:bug: Fix handling of `None` `return_predictions` in `NucleusInstanceSegmentor` CLI

Anannya30:
Problem :-
When running nucleus-instance-segmentor via CLI with AnnotationStore 
output, it crashes with:
  any(rp_ is True for rp_ in None)
  TypeError: 'NoneType' is not iterable

This happens because CLI passes return_predictions=None but line 1972 
expects a tuple.

Fix :-
Added `or (False,)` fallback:
  return_predictions = kwargs.get("return_predictions") or (False,)

Testing :-
Verified fix works by running hovernet_fast-pannuke on a real WSI 
region and successfully generating AnnotationStore output.

### DIFF
--- a/.github/workflows/conda-env-create.yml
+++ b/.github/workflows/conda-env-create.yml
@@ -11,11 +11,13 @@ on:
       - "requirements/requirements*.txt"
       - "requirements/generate_conda_env.py"
       - ".github/workflows/conda-env-create.yml"
+    branches: [ develop, master, main ]
   pull_request:
     paths:
       - "requirements/requirements*.txt"
       - "requirements/generate_conda_env.py"
       - ".github/workflows/conda-env-create.yml"
+    branches: [ develop, pre-release, master, main ]
 
 jobs:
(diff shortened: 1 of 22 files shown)


shaneahmed:
Thank you @Anannya30 . Please could you add a test to catch this error? Also as the bug is in CLI, this should be fixed in `tiatoolbox/cli/nucleus_instance_segment.py`

Anannya30:
Hi @shaneahmed, thank you for the review!

I have made both changes you requested:
1. Added the fix to `tiatoolbox/cli/nucleus_instance_segment.py` — 
   converting `None` to `(False,)` before passing to the engine.
2. Added `test_cli_default_return_predictions` in 
   `tests/engines/test_nucleus_instance_segmentor.py` to catch 
   this regression.

Please review and let me know.

shaneahmed (approved):
Thanks @Anannya30

PR #1091, opened 2026-07-02
:bug: Fix `tifffile` and `zarr` min versions on `Colab`

eshasadia:
tifffile>=2025.5.21 (e.g. 2026.4.11) imports RegularChunkGrid from zarr.core.chunk_grids, which was moved/renamed in zarr 3.2.1. This raises:

    ImportError: cannot import name 'RegularChunkGrid' from
    'zarr.core.chunk_grids'
    ValueError: zarr 3.2.1 < 3 is not supported

tifffile 2026.5.2 updated ZarrTiffStore to be compatible with zarr's current internal API and explicitly requires zarr>=3.2.0. Bump both pins in requirements.txt to avoid resolving to a broken combination.


requirements/requirements.txt
```diff
@@ -38,4 +38,4 @@ tqdm>=4.64.1
 transformers>=4.51.1
 umap-learn>=0.5.3
 wsidicom>=0.18.0
-zarr>=3.0.8
+zarr>=3.0.8,<3.2.0
```
shaneahmed:
```suggestion
zarr>=3.0.8, <3.2.0
```

shaneahmed (approved):
Thanks @eshasadia

PR #1095, opened 2026-07-09
:bug: Fix padded contours for Visualization

shaneahmed:
- This commit fixes visualization of contour overlays with padded values.
- padded values were introduced to make the contours rectangular in an effort to make np arrays compatible with zarr v3

image

fixed

image


Jiaqi-Lv (approved):
Yes the fix works, I've tested on my side.

PR #1104, opened 2026-07-27
:sparkles: Add `MC-Dropout` Uncertainty Estimation for Patch Classification

Pier45:
Adds Monte Carlo Dropout-based uncertainty quantification (epistemic + aleatoric) 
for patch classification models.

TIAToolbox's pretrained classification models (AlexNet, MobileNet, etc.) don't 
expose uncertainty estimates, only point predictions. This adds a lightweight, 
non-invasive wrapper to enable MC-Dropout sampling and uncertainty decomposition.

- `bayesian_wrapper.py`: wraps any model, injects dropout at inference if needed
- `mc_dropout.py`: runs N stochastic forward passes
- `uncertainty.py`: computes epistemic/aleatoric/total uncertainty
- Includes unit tests


tiatoolbox/models/architecture/bayesian_wrapper.py
```diff
@@ -0,0 +1,171 @@
...
+        Returns the MC-mean per-class probabilities of shape [B, C] (numpy),
+        and appends a per-batch uncertainty dict to `model.uncertainty_stats`.
+        """
+        wrapper = model  # the engine passes the wrapper as the first argument
+        base_infer = wrapper.base_model.infer_batch
+
+        samples: list[torch.Tensor] = []
+        with mc_dropout_mode(wrapper.base_model):
+            for _ in range(wrapper.n_samples):
+                sample = base_infer(wrapper.base_model, batch_data, device)
```
measty, line 150: because base_infer is just the infer_batch of the model being wrapped, when mc_dropout_mode context sets model to training mode, it will be reverted back into eval mode in any model based on CNNModel or TIMModel cause they set to eval mode inside infer_batch itself:

https://github.com/TissueImageAnalytics/tiatoolbox/blob/320ef7a3da098302e157c2b00823a8b16c1e4620/tiatoolbox/models/architecture/vanilla.py#L249

So this PR would need to solve this, likely by moving the the setting of eval from within eval_batch itself (it probably doesnt need to be there - infernce workflows should just set it once before inference)

tiatoolbox/models/architecture/bayesian_wrapper.py
```diff
@@ -0,0 +1,171 @@
...
+    The wrapped model must contain at least one Dropout layer. For pretrained
+    models that ship without Dropout (e.g. `resnet18-kather100k`,
+    `alexnet-kather100k`), pass `inject_dropout=True` (default) and a
+    `nn.Dropout` is inserted before the linear classifier automatically,
+    via `inject_dropout_before_classifier`.
+
+    After each `infer_batch` call, the per-batch uncertainty decomposition
+    (mean_probs, aleatoric, epistemic, total, mc_probs) is appended to
+    `self.uncertainty_stats` in inference order, so it can be aligned with
+    the `coordinates` array written to the engine's output `.zarr`.
```
measty, line 42: Add a minimal example of usage

measty (changes requested):
This looks useful but a few things need to be resolved:

1. the eval change (see below)
2. a bit better documentation with usage example
3. remaining test coverage

With those changes it can probably be merged

PR #1105, opened 2026-07-27
:heavy_plus_sign: Add `pysqlite3` Dependency

shaneahmed:
- Add `pysqlite3` as a dependency


shaneahmed:
not required, `conda install sqlite` is still the better option.

PR #1109, opened 2026-08-10
Manage OpenLayers frontend dependencies with npm

Demir-B:
### Summary

This PR is to add a reproducible dependency and build setup for the existing OpenLayers show-wsi frontend.

The current viewer uses pre-built copies of OpenLayers and ol-ext. The existing bundles use OpenLayers 6.9.0 and ol-ext 3.2.17, but these versions and the process used to build the frontend were not recorded in the existing setup.

This PR keeps the same dependency versions, but manages them through npm and builds the frontend with Vite. This keeps the dependency-management change separate from any future OpenLayers upgrades or viewer development.

### Changes

- Added tiatoolbox/visualization/openlayers/ containing the npm/Vite setup and frontend source.
- Added explicit OpenLayers 6.9.0 and ol-ext 3.2.17 dependencies with a lockfile.
- Moved the existing inline JavaScript and CSS from index.html into src/main.js and src/style.css.
- Vite now generates:
    static/openlayers/viewer.js
    static/openlayers/viewer.css
- Removed the previous pre-built ol.js, ol.css, ol-ext.min.js and ol-ext.min.css files.
- Added documentation for rebuilding the frontend with npm ci and npm run build.

I also tested newer OpenLayers versions during development and found differences in Zoomify tile-request behaviour. OpenLayers therefore remains at 6.9.0 in this PR to preserve the behaviour of the existing viewer.

### User and developer impact

There is no additional setup for normal TIAToolbox users. The generated JavaScript and CSS are included in the Python package, so Node.js and npm are not required to install or use show-wsi.

Developers changing the OpenLayers frontend can recreate the dependencies and build it with:

```
npm ci
npm run build
```

### Testing

- Frontend build completed successfully from a clean dependency installation (npm ci + npm run build).
- All existing tests pass and the viewer works as expected.
- Manually tested show-wsi with a WSI and multiple overlays. Overlay alignment and viewer controls function correctly with no console errors.
- Tile request behaviour matches the previous OpenLayers implementation.

### Size

The previous OpenLayers and ol-ext static files totalled approximately 1.97 MiB.

The generated viewer.js and viewer.css total approximately 0.59 MiB.
node_modules is approximately 73 MB locally, but is development-only, ignored by Git and not included in the Python package.

### Screenshots

They work the exact same just the colours are different.

Previous

old_viewer

Updated

new_viewer


shaneahmed (approved):
Thanks, @Demir-B

PR #1114, opened 2026-08-13
🔨  Add `mypy` check to `wsireader` files

Jiaqi-Lv:
## Summary
- Fix mypy errors in `wsireader/` (`types.py`, `detection.py`, `factory.py`): replace invalid `callable` types, resolve TypedDict/`**kwargs` overlaps via `WSIReaderExtraParams`, and correct related annotations.
- Bump mypy to `>=2.3` and keep local default `python_version = "3.12"` in `pyproject.toml`.
- Update the mypy GitHub Action to run on Python 3.12–3.14 with `--python-version` so each matrix job type-checks that target.


tiatoolbox/wsicore/wsireader/factory.py
```diff
@@ -28,15 +29,17 @@
         VirtualWSIReader,
         WSIReader,
     )
-    from .types import WSIReaderParams
+    from .types import WSIReaderExtraParams, WSIReaderParams
+
+PostProc = str | Callable[..., Any] | None
```
shaneahmed: Shall we move it to tiatoolbox.types as we might be using it in other modules as well.

shaneahmed (commented):
Thanks @Jiaqi-Lv This looks good. Please can you consider `PostProc`, may be we should have a generic name for models postproc and wsiread postproc?

Jiaqi-Lv:
> Thanks @Jiaqi-Lv This looks good. Please can you consider `PostProc`, may be we should have a generic name for models postproc and wsiread postproc?
 
I think that's a good idea. Shall we move it to `tiatoolbox.type_hints.py`? Don't think we have `tiatooolbox.types.py`.

shaneahmed:
> > Thanks @Jiaqi-Lv This looks good. Please can you consider `PostProc`, may be we should have a generic name for models postproc and wsiread postproc?
>  
> I think that's a good idea. Shall we move it to `tiatoolbox.type_hints.py`? Don't think we have `tiatooolbox.types.py`.

Yes, I think so.

shaneahmed (approved):
Thanks @Jiaqi-Lv

PR #1144, opened 2026-09-16
Add OpenLayers annotation controls

### LINKED ISSUE #1055: Consistent Colors for Multiple Runs
- The visualizations load with different colour schemes.
- These colours should be consistent.

Demir-B:
**Summary**

Added annotation controls and visualisation options to the experimental OpenLayers viewer.

- added an Annotations panel for viewing and editing loaded annotation types
- you can now load and control multiple annotation overlays at the same time
- if annotation colours are provided in a JSON configuration file, those colours are used first. Any annotation types without a configured colour are assigned deterministic Matplotlib colours, so the same type receives the same default colour when loaded again (this is to fix #1055)
- you can use configured annotation colours from JSON and export the current colour configuration
- annotation types can be shown/hidden and their colour and fill opacity can be changed
- added Class, Property and Class + Property display modes
  - Class colours annotations by their annotation type
  - Property colours annotations continuously using a selected numeric property (for example probability)
  - Class + Property allows one selected annotation type to use Property colouring while the remaining types keep their Class colours
- annotation types are grouped by the annotation layer they belong to
- added Show all / Hide all controls for annotation types
- clicking an annotation shows its properties and which annotation layer it came from
- added a setting to enable or disable annotation information on click
- improved how numeric property ranges are calculated so large annotation datasets can be handled without exceeding JavaScript limits
- prevented old annotation click requests from updating the viewer after a newer action

When multiple annotation overlays are loaded, Property colouring only shows a property if every overlay contains that same numeric property. 

Example with two annotation overlays loaded and coloured by Class:

image

Example of nucleus detection using Class + Property colouring, with the annotation information card open:

image


Demir-B:
**Update**

Added a few more annotation improvements:

- colour, visibility and opacity changes now apply only to the selected annotation layer/class instead of affecting matching classes in other layers
- added linked opacity
- added colour import and per-layer colour export
- added more class palettes and property colour maps
- added annotation highlighting and multiple annotation selection
- added a setting to enable multiple annotation selection
- added and updated tests

Images of the new program:

image

image

image

Demir-B:
**Update 2**

Added a few more annotation improvements:

- improved Automatic and class palette colours so loaded classes get distinct and consistent colours
- added Set2, Set3, Dark2, Accent, Paired, tab20b and tab20c class palettes
- short palettes now use colours from the other qualitative palettes when more classes are loaded instead of repeating colours
- made the final class colours consistent regardless of annotation loading order
- fixed small problem with Class + Property so nucleus detections keep their black border
- added and updated tests and documentation

